PR #8, opened 2018-05-01
Fixing bug with String deserialization where Strings weren't properly…

mew2057:
… nulled.

Fixing a defect were bad buffers result in a segmentation fault due to improper string initialization.


mew2057:
Talked with @besawn, I think it's appropriate to make a story working on a regression/unit test for the serialization engine to cover future segmentation faults. Opening a story.

PR #12, opened 2018-05-02
CSM Inventory: fix potential seg fault issue

NickyDaB:
If inventory collection API fails, then it would cause a segfault. This code checks for success. And only runs the code in case of a success, therefore avoiding a potential read of values that don't exist. eliminating a potential segfault.


mew2057:
It looks like the `CSMI_SUCCESS` if statement has been commented out. Otherwise it looks fine to me.

PR #13, opened 2018-05-02
updated csmd versioning to use major.cumulative.efix format instead of git commit

lasch:
Updated csmd to new version format and introduced back-level support for N-1 cumulative fix versions.
Also added 2 explicit git commits of the cast 1.0.0 branch that could be out there in the wild and would need to be supported.

Current limitation: the introduction of the jsrun command shifted the daemon-internal command IDs by one. Need to work on support for that.


mew2057:
+1 
Looks good to me, I think there's nothing holding back merging this.

PR #16, opened 2018-05-03
Improvement to csm_jsrun_cmd

mew2057:
- Added a new field to the input struct *jsm_path*.
- Implemented the allocation cgroup jsm placement.
- Bumped the CSM API version number to *1.1.0*

### DIFF
--- a/csmd/src/daemon/src/csmi_request_handler/CSMIAllocationDelete.cc
+++ b/csmd/src/daemon/src/csmi_request_handler/CSMIAllocationDelete.cc
@@ -25,7 +25,6 @@
 
 #include "csmi_stateful_db/CSMIStatefulDBRecvSend.h"
 
-
 #define STATE_NAME "CSMIAllocationDelete"
 
 // Use this to make changing struct names easier.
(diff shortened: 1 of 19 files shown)


mew2057:
Don't accept this has a fork bomb

lasch:
When bumping the version, you probably need to bump these guys in the main cmake file too:
```
set(BUILDID_RELEASE    1)
set(BUILDID_CUMULFIXID 0)
set(BUILDID_EFIXID     1)
```
Let me know if you want to do that while you're working on versions anyway. Otherwise, I can do that later.

mew2057:
I didn't want to bump that in this pull request, because I didn't think it was the appropriate grouping. But yes we should probably bump that too.

mew2057:
On second thought, bumping the version number in this pull request.

tgooding:
So only bump the efix ID (and in this case, it was already bumped from GA's 1.0.0 so that RPM versions would be increasing between GA and open source builds).  The cumulative version shouldn't be bumped without a larger project discussion / test cycle / etc.

lasch:
We discussed it this morning. However, it's up to others to do final decision then.
From technical perspective, there are at least 2 main reasons that would justify a cumulative version bump:
* We're introducing a whole new API 
* The version is going to be created off the master branch which i.e. not just a few cherry-picked fixes on top of CAST1.0.0.

Maybe a good time to introduce some documentation about the intended versioning process.

PR #17, opened 2018-05-03
supported CSM versions update and API cmd ID rework

lasch:
to prevent gaps in the enum range, we decided to fill the gap between regular and internal APIs with reserved IDs to prevent the internal APIs from shifting whenever a new regular API is introduced.

dropped the support for git-commit versions, i.e. the current cast 1.0.0 because of that API numbering issue.

I'm in the process of cleaning up the internal APIs too, but will do that in a separate pull request later today. Wanted to put this one out for review since it's more urgent.


csmi/src/common/include/csmi_cmds_def.h
```diff
@@ -85,3 +85,146 @@ cmd(ras_event_query_allocation)
 cmd(cgroup_login)
 cmd(jsrun_cmd)
 
+// new APIS should go here by adding the new API immediately above this line
+// and removing the first reserved_NN that appears after MAX_REGULAR?
```
besawn: I think we can drop the ? from the comment here.

lasch: I agree. the ? is not supposed to be there. Thx.

PR #20, opened 2018-05-03
Review/api checks

lasch:
With the introduction of reserved API cmd IDs, a lot of simple checks like cmd < CSM_CMD_MAX would have the potential to allow processing of the reserved IDs. To prevent this, I tried to replace this and similar checks with the new macro csmi_cmd_is_valid() to improve the filtering of invalid commands.

@mew2057 Please especially review the second commit 743faf8 because that touches your area of API serialization to make sure I didn't do anything stupid there.


fpizzano:
Looks good to me. Please Merge.

Issue #21, opened 2018-05-04
CSM Version Mismatch message not being displayed in compute log


pdlun92:
using latest daily build for master/aggregator/utility
```
ibm-csm-core-1.1.0-53.ppc64le
```
and using an old build for compute
```
ibm-csm-core-0.5.0-9237_prega.ppc64le.rpm
```

The regression bucket makes 3 checks for this test case
1. Master daemon is still running (systemctl)
2. Compute daemon is not active (systemctl)
3. Compute daemon is disconnected and `VERSION_MISMATCH` message is logged (compute log)

Checks 2 and 3 are failing.  The observed behavior is that the compute daemon stays active, and the compute log is as follows:
```
[COMPUTE]2018-05-04 08:54:33.326326       csmd::info     | CSMD: COMPUTE; Build: 302ee8da6b; Config file: /etc/ibm/csm/csm_compute.cfg
[COMPUTE]2018-05-04 08:54:33.327506       csmd::info     | Configuration: DaemonID: 29409735 using hostname: c650f03p07 PID: 120199
[COMPUTE]2018-05-04 08:54:33.328412       csmd::info     | Configuration: thread_pool_size = 4
[COMPUTE]2018-05-04 08:54:33.328629       csmd::info     | Configuration: Reading API permissions from ACL file: /etc/ibm/csm/csm_api.acl
[COMPUTE]2018-05-04 08:54:33.329219       csmd::warning  | Configuration: API=csm_jsrun_cmd NOT DEFINED!
[COMPUTE]2018-05-04 08:54:33.329325       csmd::info     | Configuration: Reading API configuration file: /etc/ibm/csm/csm_api.cfg
[COMPUTE]2018-05-04 08:54:33.329552       csmd::warning  | Configuration: API=csm_jsrun_cmd NOT DEFINED!
[COMPUTE]2018-05-04 08:54:33.333860       csmd::info     | Configuration: Found 1 buckets.
[COMPUTE]2018-05-04 08:54:33.334317       csmd::info     | CONN-HDLNG: Transition from RUNMODE_STARTED to: RUNMODE_CONFIGURED Reason: TRANSITION
[COMPUTE]2018-05-04 08:54:33.334556       csmd::info     | CONN-HDLNG: Setting up compute-aggregator connections. SingleMode=false
[COMPUTE]2018-05-04 08:54:33.334581       csmd::info     | DaemonState: Setting up 10.7.0.222:9800 as aggregator[0]
[...]

PR #22, opened 2018-05-04
fix for issue #21

lasch:
this fix will prevent the simple "disconnect because of bad msg" and turns it into "first expected msg needs to be version msg for verification".
I tested with CAST 1.0.0 daemon which gets rejected and exits with the version mismatch error.


pdlun92:
@lasch I will leave merge to you

pdlun92:
#21 Tagging issue for clarity

Issue #23, opened 2018-05-07
build error in CSMIMcast.h


adambertsch:
I was having trouble building on my branch, so I decided to build off of master first.  It looks like there's a type mismatch in the current master.  I don't know enough about the code base yet to say for certain this is the problem since I haven't chased the call path all the way yet (which is unfortunately really hard in the API due to the serialization stuff).

```
/g/g0/bertsch2/src/build/CAST/csmd/src/daemon/src/csmi_request_handler/csmi_mcast/CSMIMcast.h:94:34: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
             for( uint32_t i=0; i < _Data->num_nodes; ++i)
```

Not clear if it is intentional or not, but in different structures num_nodes is sometimes defined signed, and sometimes unsigned.  csmi_allocation_step_t is signed, while the others are all unsigned.

```
[bertsch2@butte5 CAST]$ grep -r 'int32_t num_nodes' *
csmd/src/daemon/src/csmi_request_handler/CSMIAllocationStepQueryDetails.cc:            int32_t num_nodes = output->steps[i]->num_nodes;
csmd/src/daemon/src/csmi_request_handler/CSMIJSRUNCMD.cc:        uint32_t num_nodes =  strtol(fields->data[1], nullptr, 10);
csmd/src/daemon/src/csmi_request_handler/csmi_mcast/CSMIMcastBB.h:    uint32_t num_nodes;
csmd/src/daemon/src/csmi_request_handler/csmi_mcast/CSMIMcastJSRUN.h:    uint32_t num_nodes;
csmi/include/csmi_type_wm.h:    uint32_t num_nodes; /**< Number of nodes, size of @ref compute_nodes. */
csmi/include/csmi_type_wm.h:    uint32_t num_nodes; /**< Number of nodes in the step, size of @ref compute_nodes. */
csmi/include/csmi_type_wm.h:    uint32_t num_nodes; /**< Number of nodes in allocation, size of @ref node_accounting. */
csmi/include/csmi_type_wm.h:    int32_t num_nodes; /**< Number of nodes, size of @ref compute_nodes.*/
csmi/src/wm/include/csmi_wm_type_internal.h:    uint32_t num_nodes; /**< Number of nodes, size of @ref compute_nodes. */
csmi/src/wm/include/csmi_wm_type_internal.h:    int32_t num_nodes; /**< Number of nodes, size of @ref compute_nodes.*/
[bertsch2@butte5 CAST]$ 
[...]

PR #24, opened 2018-05-08
Fixed the sign of the number of nodes in the allocation step structs.

mew2057:
Issue  #23:

Looks like the sign of the `num_nodes` for `csmi_allocation_step_t` and `csmi_allocation_step_mcast_context` was missed a while back. This fix needs to be pulled in, but JSM might break without the new headers.

@fpizzano can you contact JSM before this is merged in?


mew2057:
I should note this breaks the "promise" of versioning, but this is glaring enough of a miss/directly breaks builds that I think it's worth doing.

NickyDaB:
Hmm. Is this field related to the field "num_nodes" in the "csm_step" table of the CSM Database?

NickyDaB:
If this struct is supposed to represent a "database record" then I'm not 100% sure if we should change this value from signed to unsigned. 

The rest of this comment is under the condition that this struct is a 1:1 match to a database record for "csm_step".  

Potential concerns:  
- The database represents it as signed. If we change this value in the struct it could lead to situations with bad data representation. 
- If this is input into the database, then the API can do a check when it is called to see if the value is less than zero. Handle any errors here.
- If this is an output from the database, then I suggest to keep the data the same and present whatever the data is in the database. Even if this means a negative number. 

If this is not a struct that is supposed to match the database record, then I'm not sure if I'm feeling as strongly that it needs to stay consistent with the database.

mew2057:
Looking at this again, I think only the internal struct should be changed.

mew2057:
I think this should be good now, if you want to take a look. This shouldn't require any changes to any external users of the APIs.

mew2057:
I've been reviewing this and there is a slight concern I now have: in serialization we use the field to key array size. If it's unsigned it's technically possible that the size of the array become the unsigned maximum if a -1 were placed in the field. Maybe we need to add a secondary field used under the hood which tracks the size of the array in an unsigned manner?

NickyDaB:
> I've been reviewing this and there is a slight concern I now have: in serialization we use the field to key array size. If it's unsigned it's technically possible that the size of the array become the unsigned maximum if a -1 were placed in the field. Maybe we need to add a secondary field used under the hood which tracks the size of the array in an unsigned manner? : @mew2057 

That sounds reasonable. I sometimes take that approach on some of the inventory APIs. I usually add extra values called "arrayName_count" for those. even if there is a "num_whatever" in the database that technically represents that same number value.

Issue #15, opened 2018-05-03
libcsmpam.so inappropriately uses 'session' to perform 'account' function


adambertsch:
libcsmpam.so is currently performing account function in the session part of the module.  There are three issues to fix by separating these two functions.

1) This can cause a pam stack to initiate other session functions inappropriately when account should not have succeeded in the first place.
2) This module inappropriately prints information related to authentication during session (printing is not allowed during authentication) and this can break protocols such as rsh or scp that rely on connecting stdout/stdin to a program running on the remote host.  
3) It is desirable to be able to re-order the account pam stack separately from the session pam stack (which in libcsmpam is used to set up cgroups).  

We are currently trying to work the CAST corporate contributor agreement through our legal department.  If we are successful we can probably submit a pull request to fix these issues.

adambertsch:
I haven't quite been able to wrap my head around all of the serialization plumbing yet.  I need to split CSM_CMD_cgroup_login in to two separate CSM commands.  One that checks the active list and return success/fail to be called by pam_sm_acct_mgmt.  That is to say, everything the current CSMICGROUPLogin does, except for calling migrate_pid(), and removal of the printf.  The new class needs to perform most of the same initialization as CSMICGROUPLogin in order to call migrate_pid().  None of these modules should printf, the pam stack can output using pam_echo if it's truly desired.

adambertsch:
I was considering something like the flag approach as well.  What I couldn't convince myself was that there would never be a time when one might want to migrate pids, but not have to return an error if the current user didn't have an allocation.

Basically pseudocode like this:

account
    success if root
    fail if unable to initialize api
    success if on active list
    success if on white list
    fail

session
    migrate if on active list
    success

This way, if we call session for someone not on the active list, we don't have to fail and the logic ends up much cleaner.  accounts on the white list, or other pam module successes (that happened during account) can still cleanly pass through our session function.

Is this cleanliness worth the extra effort to do it right the first time?  I am open to being convinced that it is not especially since I'm going to have to learn a lot of stuff in order to add the APIs.  On the other hand, adding the APIs will be an excellent primer for future work.

mew2057:
Yeah I think the edit might be the path to take to follow the flag approach. Since the session code shouldn't care about the response from the migrate. I'm not 100% convinced that we need two APIs in this case simply because the flag will work well for the use case. 

[...]

PR #27, opened 2018-05-08
Add a flag for csm_cgroup_login to migrate pids on invocation or not.

adambertsch:
Add 'account' function to pam module where we do not migrate pids, but
do return success/fail.
Modify 'session' function to migrate pids, but always return pam_success
Removed printing of 'permission denied' as it is not appropriate to
print in the middle of the pam stack.

wrap struct changes for migrate_pid flag in CSM_DEVELOPMENT version macros
commit struct changes generated by the regenerate_headers.sh script

Fixes issue #15 .

This is not yet ready for merge.  Current issues are:
1) This requires pull request #24 in order to build, but with that merged in build has been confirmed.
2) The function is untested as I don't have an environment running 4/27 software to test in yet.  Hopefully later this week.
3) API version is still using CSM_DEVELOPMENT for addition to struct.  I'm not sure what the process is for the project to rev the API version, but that will need to be handled before this can be released.
4) Our CCLA is still in approval process by DoE.

### DIFF
--- a/csmd/src/daemon/src/csmi_request_handler/CSMICGROUPLogin.cc
+++ b/csmd/src/daemon/src/csmi_request_handler/CSMICGROUPLogin.cc
@@ -73,8 +73,13 @@ bool CGLoginInitState::HandleNetworkMessage(
             {
                 errCode = CSMI_SUCCESS;
 
-                std::string alloc = line.substr(delim +1);
-                migrate_pid(state_args->pid, strtoll( alloc.c_str(), nullptr, 10 ));
+                if(state_args->migrate_pid)
+                {
+                    std::string alloc = line.substr(delim +1);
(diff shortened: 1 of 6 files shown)


mew2057:
I deleted the printf in a separate merge as discussed.

adambertsch:
Merging in of this branch will also cause us to want to update the documentation for libcsmpam.  We will want people to continue to put the module in their session stack, but it will always return success in session.  That means that is no longer sufficient as a gate for access control by using 'required'.  Users will also want to put libcsmpam in 'account' with either 'required' or 'sufficient' depending on their pam stack configuration in order to allow users access when they have active jobs on the node.

If a user passes the pam stack on a node where they don't have an allocation, but an allocation exists that uses core isolation, they will end up in the system cgroup.

I can write the documentation if there's an appropriate document to update in git, or I can work with a document author at IBM.

fpizzano:
CCLA signed for @adambertsch

Issue #31, opened 2018-05-09
csm_allocation_delete is not capturing the correct information on failure


fpizzano:
csm_allocation_delete is not capturing the correct information on faulire:

```
# /opt/ibm/csm/bin/csm_allocation_query_details -a 28
---
allocation_id:                  28
primary_job_id:                 373
secondary_job_id:               0
num_nodes:                      18
compute_nodes:
  f0n07:
   power_cap:                   3050
   power_shifting_ratio:        100
  f0n08:
   power_cap:                   3050
   power_shifting_ratio:        100
  f0n09:
   power_cap:                   3050
   power_shifting_ratio:        100
  f0n11:
.....
```

```
csmdb=# select * from csm_allocation_node_history where allocation_id = '28' and node_name = 'f0n07';
        history_time        | allocation_id | node_name |  state  | shared | energy | gpfs_read | gpfs_write |   i
b_tx    |   ib_rx    | power_cap | power_shifting_ratio | power_cap_hit | gpu_usage | gpu_energy | cpu_usage | mem
ory_usage_max | archive_history_time
----------------------------+---------------+-----------+---------+--------+--------+-----------+------------+----
--------+------------+-----------+----------------------+---------------+-----------+------------+-----------+----
--------------+----------------------
 2018-05-08 13:29:59.218207 |            28 | f0n07     | running | f      |      0 |  17137664 |       8192 |  39
6811784 |  401237665 |      3050 |                  100 |             0 |        -1 |            |           |
              |
 2018-05-08 13:30:32.356493 |            28 | f0n07     | running | f      |      0 | -17137664 |      -8192 | -39
6811784 | -401237665 |      3050 |                  100 |             0 |        -1 |            |         0 |
            0 |
 2018-05-08 13:30:32.356493 |            28 | f0n07     | failed  | f      |      0 | -17137664 |      -8192 | -39
6811784 | -401237665 |      3050 |                  100 |             0 |        -1 |            |         0 |
            0 |
(3 rows)

csmdb=#
```

mew2057:
I have a candidate fix that needs to be tested on a system with the data aggregators enabled, making a pull request now.

pdlun92:
@mew2057 @fpizzano ok to close this one?

mew2057:
I think we should be able to since the fix has been merged/tested.

fpizzano:
Looks good running:

```
# rpm -qa | grep ibm-csm
ibm-csm-api-1.1.1-183.ppc64le
[...]

PR #34, opened 2018-05-09
Allocation Delete failures now save metrics if possible.

mew2057:
Candidate fix for Issue #31:

Implemented a new mechanism for handling softer failures in multicast code. Allocation Delete failures will now maintain the metrics after failing.

Added a new field `_ErrorMsg` to the multicast properties template class to facilitate this mechanism.


csmd/src/daemon/src/csmi_request_handler/CSMIAllocationAgentUpdateState.cc
```diff
@@ -117,15 +117,14 @@ bool AllocationAgentUpdateState::HandleNetworkMessage(
     }
     
     // If the initialization or reversion was successful reply to the master daemon.
-    if ( success )
+    if ( success || response->error_code != CSMI_SUCCESS )
```
NickyDaB: if success or not success? Only line i'm confused about. otherwise nothing jumps out as crazy. but I don't know enough about this part of the code to offer any real comment of substance. 

explain this to me then I will be ready to approve.

mew2057: This section effectively says that if the `RevertNode` or `InitNode` returned true (meaning the function completed in a good execution) respond. Additionally, if the response has an error code, meaning the `RevertNode` executed, but it had problems, we want to respond with the data enriched path.

By having both options in the if statement we capture cases where there's no data enrichment, but a failure was found in one of the functions or before the functions got called (`response->error_code` defaults to `CSMI_SUCCESS`). Additionally, we don't skip data enriched payloads just because the function returned an error.

NickyDaB: ight

NickyDaB:
Looks like everyone approves. So, I guess you are cleared for pull.

PR #36, opened 2018-05-09
Enabling python for gcc.

mew2057:
Python generation was previously disabled in gcc due to a misunderstanding of the pragma message. This change will now enable the compilation of the CSM python libraries (first pass).

### DIFF
--- a/csmi/src/CMakeLists.txt
+++ b/csmi/src/CMakeLists.txt
@@ -51,69 +51,66 @@ install(TARGETS csmi_serialization COMPONENT csm-core DESTINATION csm/lib)
 
 add_subdirectory(launch/src)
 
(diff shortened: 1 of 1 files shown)


mew2057:
@fpizzano We should probably merge this for the next release so users can play with the first release version of the python bindings.

tgooding:
fyi- this merge broke the (gcc) build.

mew2057:
Looks like the problem was there were new comments that included quotes that were automatically pulled in that hadn't been accounted for in branch this was built against. I have a quick fix that I'm testing now and pushing to master.

The long and short of it was the meta program was not designed to handle double quotes, but it looks like some of the inventory structs were given comments with double quotes.

PR #42, opened 2018-05-10
BB fixes for various issues identified by clang analyzer

tgooding:


tgooding:
@dlherms-ibm :  can you double-check my changes?  In particular, i'm worried about a few:

Commit 10f6bb1cb4685558e1e932dd717c2dfc03cd695b seems like the rc=-1 should error out.  But I left the failure as suppressed like it was beforehand.  

Commit b3514b5858020acc085a0d4693668980a12dec55 sets the local scoped fd at the end.  Seems like it should have been the classes' lockfd instead.

tgooding:
And also:
Commit df72f9acae39780e48c6f8416b1c00990dc317b9 - there are 2 rc variables at different scopes.  Seems like we should only use 1.  As it stands, the rc in the inner scope is forgotten.

dlherms-ibm:
The first seven commits look good to me...  Nit...  You could keep the 'return;' statement in LVUtils.cc

Checking the rest now...

bb/src/BBTagInfo.cc
```diff
@@ -143,19 +143,12 @@ int BBTagInfo::addTransferDef(const std::string& pConnectionName, const LVKey* p
...
-                rc = 0;
-            }
-            else
-            {
-                rc = -1;
-                errorText << "BBTagInfo::addTransferDef: For " << *pLVKey << ", handle " << pHandle << ", contribid " << pContribId \
-                          << " was already known to the cross-bbServer metadata.";
-                LOG_ERROR(errorText);
-            }
+//          rc = -1;
```
dlherms-ibm: We still want to set rc = -1 here...

tgooding: But the subsequent line sets rc = parts.addTransferDef()....

dlherms-ibm: Yep...  Correct...

It was wrong before also...  We should set rc = -1 if we are not restarting the transfer definition and return before calling parts.addTransferDef().  As it is now, the addTransferDef() call should also return -1, which makes it fail as it should. 

Actually, thinking about it again, might be best to leave it as it was originally and just add a check for a (!rc) before calling parts.addTransferDef().  Otherwise, in the restart case, we leave the rc as 1 and then call parts.addTransferDef() to overwrite rc.  That is a little odd to me...

tgooding: I'm not sure I follow where you landed, can you checkin what you'd prefer?

tgooding:
Runs clean now, merging.

PR #45, opened 2018-05-10
Feature csm api cluster query state

NickyDaB:
Adds in a new CSM API. **csm_cluster_query_state**

This API will query all the nodes in your cluster and tell you:
- when they were last updated
- what state they are in
- what type of node they are
- how many allocations they are currently participating in

if that number is greater than 0 then also show:
- the ids of those allocations 
- the states of that allocation (ie: staging-in, running, staging-out)
- and if that allocation is shared (true or false)

Can also filter the query by a few options to refine the data returned.


csmd/src/daemon/src/csmi_request_handler/CSMIClusterQueryState.cc
```diff
@@ -0,0 +1,246 @@
...
+		}
+		stmt.append("GROUP BY n.node_name ");
+		if(input->num_allocs >= 0)
+		{
+			stmt.append("HAVING ( COUNT(an.allocation_id) = $");
+			stmt.append(std::to_string(++SQLparameterCount));
+			stmt.append("::int ) ");
+		}
+		stmt.append("ORDER BY ");
+		switch (input->order_by)
```
mew2057: Maybe we can convert this into a function or hash in the future so it can be reused in other APIs?

NickyDaB: The order by?

mew2057: I mean the switch statement, it might be nice to formalize that as a mechanism for the future so we can easily drop it into other code.

NickyDaB: oh, ok. Yes, we should talk about that in the future. I haven't really formalized any of the order by patterns yet. Some APIs don't have any, some have 2, and some have a lot like here.

PR #49, opened 2018-05-14
selecting aggregators for multicast msgs

lasch:
This pull request will add a feature that allows the master to match and select the aggregators that are responsible for the compute nodes listed in a multicast message.
It will send to both primary and secondary aggregator in case any failover has not made it to the master yet.
For any multicast that spans more than 1024 nodes, it will send to all aggregators because the overhead of parsing and matching the node list, would outweigh the overhead of sending a message to all aggregators. With further tests and optimization in the future, this threshold can be adjusted.
The changes also include an update to the up/down reporting of compute nodes between aggregator and master: up to now, a quick cycle of disconnect+connect of a compute node might be accumulated into a no-op because the down and up events are consolidated. This was changed into an actual log of up and down events including the connection sequence number so that no transient down+up can be missed.

Before considering a merge, this needs thorough testing on a system with the following requirements (this is likely one for @fpizzano and his test system):
* needs more than 2 aggregators
* several overlapping sets of compute nodes configured with primary and secondary aggregators

The tests should involve an API that uses multicasts (e.g. allocation_create/delete)
* Test different subsets of compute nodes so that the selection mechanism would kick in.
* Every possible subset of compute nodes must succeed
* Maybe confirm that only a subset of aggregators is selected by setting the master loglevel: csmd::debug and look for lines like:
 ```
 2018-05-14 14:13:19.887212       csmd::debug    | NETMGR: Resolved ABSTRACT_ADDRESS_AGGREGATOR to 2 destinations.
 ```
 where the 2 destinations need to be at least the required minimum of aggregators for the given compute set. With the whole system working, this number can never be less than 2.
* Tests should also cover the fast down+up cycle of compute nodes to see if all events get recognized at the master and corresponding RAS events are created. This probably can only be achieved by cable pull tests - maybe also by SIGKILL+restart of compute daemons.


lasch:
@mew2057 this pull request could use the sorted node list you worked on at one point. I've built in a check and a sort in case the input is not sorted, however that would add unnecessary overhead.

PR #55, opened 2018-05-17
Updated csm_ras_event_create backend to use parameterized SQL for DB access

besawn:


NickyDaB:
Looks good. In the future, when you work on your next feature, I don't think it's a good practice to pull in from the main master into your feature branch, unless you really need those new commits from master in your feature branch. Other than that I don't have any comments. Approving.

NickyDaB:
@besawn - It looks like all reviewers have given the "thumbs up" so you can pull this into the main project whenever you want. 

Thanks for embracing the pull request practice and procedure.

PR #59, opened 2018-05-18
Cgroup Debug/Error Logging

mew2057:
Modified the error logging for cgroups. Now in the event of a failure a list of the pids that can't be killed will be logged to the error utility. 

Moved a warning error message down to debug to reduce confusion.

### DIFF
--- a/csmd/src/daemon/src/csmi_request_handler/helpers/cgroup.cc
+++ b/csmd/src/daemon/src/csmi_request_handler/helpers/cgroup.cc
@@ -697,10 +697,11 @@ void CGroup::DeleteCGroup(
         // If the cgroup fails to kill all of the tasks throw an exception.
         if (tasksFound > 0 )
         {
+            KillTasks( groupPath, true );
             std::string error = 
                 "DeleteCGroup; Unable to remove " + std::to_string(tasksFound) +
                 " tasks after " +  std::to_string(killAttempts) + 
-                " attempts in the " + groupPath + " cgroup.";
+                " attempts in the " + groupPath + " cgroup;";
             throw CSMHandlerException( error, CSMERR_CGROUP_FAIL );
         }
     }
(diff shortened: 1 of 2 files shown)


csmd/src/daemon/src/csmi_request_handler/helpers/cgroup.cc
```diff
@@ -697,10 +697,11 @@ void CGroup::DeleteCGroup(
         // If the cgroup fails to kill all of the tasks throw an exception.
         if (tasksFound > 0 )
         {
+            KillTasks( groupPath, true );
             std::string error = 
                 "DeleteCGroup; Unable to remove " + std::to_string(tasksFound) +
                 " tasks after " +  std::to_string(killAttempts) + 
-                " attempts in the " + groupPath + " cgroup.";
+                " attempts in the " + groupPath + " cgroup;";
```
NickyDaB: Reason to end the sentence in a semi colon instead of the period? Just wondering.

mew2057: I've been trying to delimit log messages with semi colons when possible if it's being added to the error message to make parsing error messages a little simpler.

NickyDaB: Ah. ok. Is this for all log messages, only CSM APIs, only big data store, etc etc?

pdlun92:
Ran a unit test on this and was able to observe the new error message in the compute log.  However I have a couple of comments.  They may be related to my test case, which I found accidentally but was able to reproduce the failure case consistently.  My test case is as follows

1) create allocation
2) run jsrun_cmd using that allocation
3) delete the allocation

These commands are all run immediately after each other.  I think it may be a race condition that is forcing the error, though I am not sure.  May be worth noting that I can reliably recreate the error condition using that test case on the FVT P8 nodes, while the P9 nodes seem to handle it fine.

But my main question is regarding additional error handling (again, I do see the compute message in the log, so within the scope of this pull request I think the code can be approved)... The observed behavior is that this message is logged, the allocation is deleted from the database, but the cgroup directory for the allocation sticks around, and after the test case is run the tasks list for that allocation appears to be empty.  

I think I need to have some discussion on this, whether A) my test case is valid and B) the observed behaviors are what we want

PR #60, opened 2018-05-21
enable ctrl-cmd to reset primary/secondary connection status according to config

lasch:
This adds a new feature/option to csm_ctrl_cmd:
```
csm_ctrl_cmd --agg.reset
```
This requests a compute daemon to reset it's primary/secondary connections in case the primary/secondary had changed because of aggregator fails/restarts etc.

A few details for whoever is interested:
* new internal ctrl msg between compute and agg to tell a primary agg to reset connection status to secondary - was required because both aggregators stay up and therefore this is not possible to be processed like a failover
* added a system event (RESET_AGG) to allow API handlers to trigger system events (extension of former JOB_START/STOP event handling); this makes it easier to process additional system events in the future
* existing ctrl-cmd extended to trigger the RESET_AGG system event; note: as with the existing ctrl-cmd this needs to be called on the same node as the compute daemon that is to be reset.

For testing, I'll create an issue that describes how to test.

(Apologies for mixing in an extra commit that fixes a typo in one of the files that has additional changes. I could have buried it within the other commits, but I prefer to keep things contained. ;-) )

### DIFF
--- a/csmd/src/daemon/src/csm_daemon_network_manager.cc
+++ b/csmd/src/daemon/src/csm_daemon_network_manager.cc
@@ -409,6 +409,16 @@ bool csm::daemon::EventManagerNetwork::ProcessNetCtlEvents()
             break;
         }
         break;
+      case csm::network::NET_CTL_RESET:  // compute tells us that it switched to a different primary aggregator
+        switch( _Config->GetRole() )
+        {
+          case CSM_DAEMON_ROLE_AGGREGATOR:
+            _DaemonState->SetConnectionTypeEP( info_itr->_Address, csm::daemon::ConnectionType::SECONDARY );
+            break;
+          default:
+            break;
+        }
+        break;
       case csm::network::NET_CTL_TIMEOUT:
       {
         csm::daemon::MessageContextContainer_sptr msgCtx = _MessageControl.FindMsgAndCtx( info_itr->_MsgId, true );
--- a/csmd/src/daemon/src/csm_daemon_network_manager.cc
+++ b/csmd/src/daemon/src/csm_daemon_network_manager.cc
@@ -416,6 +416,7 @@ bool csm::daemon::EventManagerNetwork::ProcessNetCtlEvents()
             _DaemonState->SetConnectionTypeEP( info_itr->_Address, csm::daemon::ConnectionType::SECONDARY );
             break;
           default:
+            CSMLOG( csmd, info ) << "NET_CTL_RESET event at non-aggregator daemon. Ignoring...";
             break;
         }
         break;
--- a/csmnet/include/csm_network_header.h
+++ b/csmnet/include/csm_network_header.h
@@ -69,6 +69,13 @@
 #define CSM_FAILOVER_MSG "FAIL0VER"
 
 
+/** @def CSM_RESET_MSG
+ * @brief String to be sent between compute and aggregator to
+ *  signal a reset of primary connection to secondary
+ */
+#define CSM_RESET_MSG "R3S3T"
+
+
 /** @def CSM_RESTART_MSG
  * @brief String to be sent between compute and aggregator to
  * signal a restarted or "fresh from disconnect" state
--- a/csmd/src/daemon/include/connection_handling.h
+++ b/csmd/src/daemon/include/connection_handling.h
@@ -241,9 +241,9 @@ protected:
   int SecondaryUp();
   virtual int CreatePrimary();
   void QueueFailoverMsg( csm::network::Address_sptr addr );
+  void QueueResetMsg( csm::network::Address_sptr addr );
   int UpTransitionAndFailover( const csm::network::Address_sptr i_Addr,
                                const bool i_Failover );
-
 private:
   int CreateSecondary();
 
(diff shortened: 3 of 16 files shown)


NickyDaB:
Hey @lasch when you open your next issue, don't forget to add on the 2 reviewers (1 dev and 1 regression) and assign yourself to eventually pull in the issue. 

I've added myself and @williammorrison2 as reviewers, and assigned you to the main issue.

lasch:
@NickyDaB I'll take care of the assignment to myself in the future. However, I don't want to force someone specific to review something where I don't know if that person is the right person. So, if I know who to point out as reviewer, I'll set the reviewer, otherwise, I'll leave that open (unless we have a default person that should be added to review everything). Can you describe how you came up with Bill and yourself to be selected reviewers as opposed to e.g. Pat and John/Nate?

NickyDaB:
@lasch 



We established some default reviewers last Thursday during the call. We said that the "default" reviewers are:
- 1 developer (to give the code a sanity check)
- 1 regression checker (to make sure the code passes regression testing)

For now those definitions are:
A developer can be: John, Lars, Nate, Nick (developers of CSM)
A regressor can be: Pat or Bill (someone who works with regression testing for CSM)

Doesn't exactly matter who, as long as its someone to look it over. So for now, you can just rotate through the list. 

with the only exception is that you can not select yourself as a reviewer, as that would be counter to the mission. 

You can have more should you want. ie you can select 2 developer reviewers. (John and Nate), but you must still meet the min 1 dev and 1 reg req. 

You can have others in addition to our 2 minimum catagory. ie, if I make a new burst buffer CSM API, then I should also probably ask tom gooding to review it in addition to our 1 dev and 1 regressor.

csmd/src/daemon/src/csm_daemon_network_manager.cc
```diff
@@ -409,6 +409,16 @@ bool csm::daemon::EventManagerNetwork::ProcessNetCtlEvents()
             break;
         }
         break;
+      case csm::network::NET_CTL_RESET:  // compute tells us that it switched to a different primary aggregator
+        switch( _Config->GetRole() )
+        {
+          case CSM_DAEMON_ROLE_AGGREGATOR:
+            _DaemonState->SetConnectionTypeEP( info_itr->_Address, csm::daemon::ConnectionType::SECONDARY );
+            break;
+          default:
```
NickyDaB: Your : `switch( _Config->GetRole() )` statement has a default case of nothing?

It makes sense to me that you want to have a switch to account for different daemon roles? correct?

why do you leave the default case blank?

Should you print out some error if it does not match the `CSM_DAEMON_ROLE_AGGREGATOR` case?

I'm wondering what the thought process here was, so I can get on the same page.

lasch:
@NickyDaB Haha... No idea where I was when that was agreed on. Anyway, it took a long text to describe that the reviewer selection is completely arbitrary if I'm the only one working on that piece of code. I'll follow that process until we hopefully come up with a better one soon.. ;-)

csmnet/include/csm_network_header.h
```diff
@@ -69,6 +69,13 @@
 #define CSM_FAILOVER_MSG "FAIL0VER"
 
 
+/** @def CSM_RESET_MSG
+ * @brief String to be sent between compute and aggregator to
+ *  signal a reset of primary connection to secondary
+ */
+#define CSM_RESET_MSG "R3S3T"
```
NickyDaB: Just wondering: Why do we put numbers into these keywords? like: `"FAIL0VER"` and `"R3S3T"` Does it make the message more secure?

csmd/src/daemon/src/csm_daemon_network_manager.cc
lasch: It's blank because at current status of code, there's nothing to be done on other daemon roles in case they receive a RESET event.
In this case, it copies the style from the handling of the FAILOVER event.
The handling of other events should probably follow that style sooner or later too.
If this message hits another daemon, there's no harm that would demand a warning at the moment.

csmnet/include/csm_network_header.h
lasch: That has no impact at all on security.
It's just more fun ;-) and this way one can separate the string in the message from other strings in the log that say FAILOVER, or RESET, or DISCONNECT. Other than that, the strings for these ctrl messages are completely arbitrary and can be set to whatever we would want in that single place in the network header file.

csmd/src/daemon/src/csm_daemon_network_manager.cc
NickyDaB: ok. What about a debug, trace, or log that says something like "This daemon does not currently support reset." This way the user can get some feedback. It would also show us in the logs that someone is calling reset on a daemon that they shouldn't be.

csmnet/include/csm_network_header.h
NickyDaB: oh ok.

csmd/src/daemon/src/csm_daemon_network_manager.cc
lasch: Yeah, I can add that.

NickyDaB: cool. thanks.

lasch: pushed a commit with a message for non-aggregator daemons.

PR #64, opened 2018-05-21
Cgroup improvement, added additional task migration

mew2057:
Implemented a new feature which migrates tasks to the non cpuset system cgroup when the allocation is created. The following cgroups should all have the same number of pids in the `tasks` file after the allocation is created:

* `/sys/fs/cgroup/cpuset/csm_system`
* `/sys/fs/cgroup/memory/csm_system`
* `/sys/fs/cgroup/devices/csm_system`
* `/sys/fs/cgroup/cpuacct/csm_system`

Cgroup tests should be modified to check this in the near future.


mew2057:
@pdlun92 This should be ready for a second unit test.

pdlun92:
Looks good with the exception of some nvidia-irq tasks that can't be migrated to `cpuacct` and some tasks related to pam config that can't be migrated to `devices` `memory` or `cpuacct`.  However these changes seem acceptable.  Approving

Issue #65, opened 2018-05-21
csmd utility daemon SEGFAULT on NetworkManager


fpizzano:
Utility daemon SEGFAULT on NetworkManager

```
[UTL]2018-05-17 11:21:38.121519       csmd::info     | CSMD: UTILITY; Build: 1.1.0; Config file: /etc/ibm/csm/csm_utility.cfg
```

```
[UTL]2018-05-20 21:57:45.552520     csmapi::info     | CSM_CMD_node_resources_query_all[1999289472]; Client Recv; PID: 85848; UID:0; GID:0
[UTL]2018-05-20 21:57:45.552602     csmnet::info     | Deleting known client address: tmp/csmi0.85848.20971
[UTL]2018-05-20 21:57:45.552620     csmnet::warning  | Found disconnecting endpoint with pending Accept() of client tmp/csmi0.85848.20971
[UTL]2018-05-20 21:57:45.552689     csmnet::info     | Sync:Active:Endpoint removal: 192.168.75.124:9816  remaining=1
[UTL]2018-05-20 21:57:45.553052       csmd::info     | DAEMONSTATE: DISCONNECTEP: 192.168.75.124:9816
[UTL]2018-05-20 21:57:45.553073       csmd::info     | DAEMONSTATE: UpdateEPStatus(): Status of connection to: 192.168.75.124:9816 change to RUNMODE_DISCONNECTED
[UTL]2018-05-20 21:57:45.553091       csmd::info     | CONN-HDLNG: Removing listening endpoint type:CSM_NETWORK_TYPE_LOCAL addr=/run/csmd.sock
[UTL]2018-05-20 21:57:45.553110     csmnet::info     | ConnectionHandling::TakeDependentDown:Endpoint removal: /run/csmd.sock  remaining=0
[UTL]2018-05-20 21:57:45.553214       csmd::info     | CONN-HDLNG: Transition from RUNMODE_READY_RUNNING to: RUNMODE_DISCONNECTED Reason: ERROR
[UTL]2018-05-20 21:57:45.554662       csmd::critical | Caught segfault when accessing data at address: 0x78
[UTL]2018-05-20 21:57:45.554693       csmd::critical | Instruction Address from where the crash happened: 0x109ad08c
[UTL]2018-05-20 21:57:45.554709       csmd::critical | SEGFAULT: thread_id = 7fff8174f040 handler_name = NetworkManager
```

fpizzano:
[...]

PR #66, opened 2018-05-21
candidate fix for #65

lasch:
This should fix the crash.
However, I have not been able to recreate.It could be hard to reproduce. See comment in issue #65


lasch:
Are we waiting for a test case to show it's fixed before we merge this? I mean it's marked sev1, so I better spend my whole life on this before it's fixed.
Should we merge and keep track of trouble and progress in issue #65 ?
Otherwise this fix wouldn't make it into mainline anytime soon.

fpizzano:
I do agree to merge and track progress in issue #65.

PR #87, opened 2018-06-04
Feature improvement/switch query

NickyDaB:
Improve features for CSM API: **csm_switch_attributes_query**

Changes include:
- Added in an 'order_by' option into the input of the API. (This lets a user order the results by supplied input)
- Removed the minimum '1 filter argument' restriction on the API. (If the user doesn't supply one of the optional filter arguments, then the API returns all switch records in the system.)
- Added in a "JSON" output option to the CMD line interface of this API.
- Added in a "format" output option to the CMD line interface of this API. (first pass at integrating the CSM format tool)


NickyDaB:
Thanks @williammorrison2

NickyDaB:
Thanks @besawn

NickyDaB:
pulling this request into development

PR #95, opened 2018-06-07
csm: fix client-to-daemon large message protocol

lasch:
We noticed that the large-msg protocol for messages from clients to daemon was not functioning as expected. It caused the daemon to receive several invalid messages and the client to get disconnected.
This PR fixes the protocol.

@fpizzano This might be a candidate for a hotfix since it might cause trouble for anything that tries to send CSM API requests larger than 128k.

Will open an issue to add a regression test.

### DIFF
--- a/csmd/src/daemon/src/csm_daemon_network_manager.cc
+++ b/csmd/src/daemon/src/csm_daemon_network_manager.cc
@@ -577,7 +577,15 @@ bool csm::daemon::EventManagerNetwork::EndpointSendActivity()
   }
   if( ! data_sptr->_Msg.Validate() )
   {
-    CSMLOG(csmd, error) << "BUG: Found invalid outbound message! Returning send error to handler.";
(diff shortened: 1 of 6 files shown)


fpizzano:
For the records: 

# error_inject -m 16777120    -> Work

# error_inject -m 16777121    -> Fail

lasch:
With the latest few commits, the limit slightly shifts to 16777143.
Without these commits there was a problem with a truncated msg that would cause the receiving client to time out. Instead, the daemon shouldn't even attempt to send a msg that would get truncated.
This has been reviewed and fixed. @pdlun92 I'd suggest to retry the regression.
@fpizzano let me know if you want to experiment with the latest set of commits before we merge.

pdlun92:
@lasch Ran regression on the new code, passed again.  Additionally the unit tests are still behaving as expected

PR #107, opened 2018-06-12
csm_db_update: added more features and functions to the archiving and…

williammorrison2:
… deletion scripts


mew2057:
@williammorrison2 Hey I missed `csm_db_history_archive.sh` in the CMAKE file, can you add it in since you've got the files in flight?

PR #119, opened 2018-06-18
hostname and command configuration improvements

lasch:
This pull request contains a small set of adjustments to handle configured host names and introduce 'NONE' as an option to ignore a host name setting (currently for secondary aggregator or BDS).

For example, a 'clean' single-aggregator config would look like this:
```
            "aggregatorB" :  {
                "host": "NONE",
                ...
```
It also enables to use comments in csm_api.cfg: any command that contains a # will be ignored. See example:

```
{
   "# comment" : "more comment",
   "also com#ment" : "possible but not recommended",
   "csm_allocation_create" : 500,
   "csm_allocation_delete" : 60,
   ...
```

We might want test cases that cover the 3 cases:
* valid entry to enable the config
* invalid entry that causes a warning
* NONE option to explicitly disable

### DIFF
--- a/csmconf/csm_aggregator.cfg
+++ b/csmconf/csm_aggregator.cfg
@@ -11,7 +11,7 @@
             "format"                    :   "[AGG]%TimeStamp% %SubComponent%::%Severity% | %Message%",
             "consoleLog"                :   false,
             "fileLog"                   :   "/var/log/ibm/csm/csm_aggregator.log",
-            "__rotationSize_comment_1"  :   "Maximum size (in bytes) of the log file, 1000000000 bytes is ~1GB",
+            "#rotationSize_comment_1"  :   "Maximum size (in bytes) of the log file, 1000000000 bytes is ~1GB",
             "rotationSize"              :   1000000000,
             "default_sev"               :   "warning",
             "csmdb"                     :   "info",
@@ -27,8 +27,8 @@
             "heartbeat_interval" : 15,
             "local_client_listen" :
             {
-                "__socket_comment_1"  : "Use /run/csmd.sock by default.",
-                "__socket_comment_2"  : "Use /run/csmd_agg.sock when running a Master and Aggregator daemon in the same node.",
+                "#socket_comment_1"  : "Use /run/csmd.sock by default.",
+                "#socket_comment_2"  : "Use /run/csmd_agg.sock when running a Master and Aggregator daemon in the same node.",
                 "socket"              : "/run/csmd.sock",
                 "permissions"         : 777,
                 "group"               : ""
--- a/csmconf/csm_api.cfg
+++ b/csmconf/csm_api.cfg
@@ -1,4 +1,5 @@
 {
+   "#comment_1" : "This will be ignored",
    "csm_allocation_create" : 120,
    "csm_allocation_delete" : 120,
    "csm_allocation_update_state" : 120,
--- a/csmconf/csm_compute.cfg
+++ b/csmconf/csm_compute.cfg
@@ -12,7 +12,7 @@
             "consoleLog"                :   false,
             "sysLog"                    :   true,
             "fileLog"                   :   "/var/log/ibm/csm/csm_compute.log",
-            "__rotationSize_comment_1"  :   "Maximum size (in bytes) of the log file, 100000000 bytes is ~100MB",
+            "#rotationSize_comment_1"  :   "Maximum size (in bytes) of the log file, 100000000 bytes is ~100MB",
             "rotationSize"              :   100000000,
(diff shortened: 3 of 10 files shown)


besawn:
In the main CSM configuration files (csmconf/csm_aggregator.cfg, for example), we use this convention for comments:

```
            "local_client_listen" :
            {
                "__socket_comment_1"  : "Use /run/csmd.sock by default.",
                "__socket_comment_2"  : "Use /run/csmd_agg.sock when running a Master and Aggregator daemon in the same node.",
                "socket"              : "/run/csmd.sock",
                "permissions"         : 777,
                "group"               : ""
            },
```
It seems like it would a good idea to adopt a consistent convention across all of the various configuration files. Using # as a comment character seems like an improvement over __ as we are using __ for various configuration eye catchers.

However, do you know if something like this will be able to be parsed by boost property tree:
```
            "local_client_listen" :
            {
                "#"  : "Use /run/csmd.sock by default.",
                "#"  : "Use /run/csmd_agg.sock when running a Master and Aggregator daemon in the same node.",
                "socket"              : "/run/csmd.sock",
                "permissions"         : 777,
                "group"               : ""
            },
```
Or will it fail with something like a duplicate key error, with the key "#" being reported as duplicated?

Something like this also seems fine if there is a duplicate key limitation:
```
            "local_client_listen" :
            {
                "#socket_comment_1"  : "Use /run/csmd.sock by default.",
                "#socket_comment_2"  : "Use /run/csmd_agg.sock when running a Master and Aggregator daemon in the same node.",
                "socket"              : "/run/csmd.sock",
                "permissions"         : 777,
                "group"               : ""
            },
```

Do you know if there is a duplicate key limitation?

Do you think we should extend this "#" handling to the main CSM configuration files, too?

lasch:
Very good points @besawn . 
Regarding the # vs __: in the daemon config files, we can use anything we want because there's no parsing of all keys in that file. However, I agree, we should use only one type of comment for consistency.  Should we stick to '#' and adjust the config templates? My suggestion would be to move the comment keys to use '#'. However, I'd probably keep the __MASTER__ and similar items as is since they are more placeholders or variables than comments.

Regarding duplicate "#": I saw that the json standard is no enforcing any behavior. Although, certain implementations might do that. I ran with multiple "#"-only keys and there was no complaint about duplicate keys from boost while parsing. I'd say, we don't bother preventing duplicate keys because it's eventually up to the user to provide a json standard compliant config file. The daemon should take care of parsing errors and provide proper guidance through error msgs in case the parser rejects a file. However, our examples and templates should probably not encourage duplicates.

besawn:
@lasch 
I agree with the suggestion to modify the config templates to adopt this same approach using '#'.
I also agree that we should continue to use __ for the placeholders.

It is good that duplicates are tolerated by boost while parsing, and I agree that we should try not to encourage duplicates in any example we provide. I think this approach would direct us toward something like this option from above:
```
            "local_client_listen" :
            {
                "#socket_comment_1"  : "Use /run/csmd.sock by default.",
                "#socket_comment_2"  : "Use /run/csmd_agg.sock when running a Master and Aggregator daemon in the same node.",
                "socket"              : "/run/csmd.sock",
                "permissions"         : 777,
                "group"               : ""
            },
```

lasch:
Sounds good. I'll update the template files accordingly before merging, so we can keep these changes contained within one merge.

pdlun92:
I also ran regression on this build, along with the unit tests.  All proper behaviors and error/warning/info log messages were observed.  I do have some quick cases that I just wanted to clear up to make sure they are working as intended

1. Aggregator A: *Valid* Aggregator B: *Invalid* (valid syntax but unknown hostname) 
error message in log and daemon exits
```
csmd::error    | Configuration: AggregatorB: Unable to resolve hostname: c650f03p14
```
2. Aggregator A: *Valid* Aggregator B: *Valid* but both are configured to the same host
error message in log and daemon exits
```
csmd::error    | Both configured aggregators are on the same host. csmd build is configured to run one aggregator per node only.
```

These clearly seem to be intended behaviors but I just want to explicitly confirm

lasch:
@pdlun92

 1.  If you configure aggregatorB as a valid hostname string but pointing to an unknown host, the daemon will keep trying to connect every ~5 seconds (same should apply if aggregatorA is set up that way). So, the observed behavior is as expected. It would be very unwise to set it up that way, though.

 2. short answer: is correct behavior. Long answer: >1 csmd of the same role on the same host is not supported. You'd need a special compile flag to enable operation with 2 aggregators on the same host (and it's only intended for debugging/dev because it has side-effects for restarted daemons where each restart/reconnect would be considered as a new node)

PR #124, opened 2018-06-19
CSM FVT test cases for large msg protocols

### LINKED ISSUE #96: large mgs protocol tests
We found out that the large-msg protocol for the direction from client to daemon was broken. We need to add a test case that causes a client to send a large msg to a daemon.
Pull request #95 fixes the protocol.

The simplest option is to use the error_inject tool, however, it's only in the build tree and not included in the RPM right now:
```
error_inject -m <msg_size>
```
whith msg_size to cover 3 cases:
 - m < 128k
 - 128k < m < 2x128k
 - m > 2x128k
Reason for the 2x category is that the case that made us find the issue was in that range and it caused the operation to successfully complete on the daemon side but then followed by disconnects and error logging. A case with > 2x would have failed completely.

Another option could be through the IB switch API which sends the content of a file to the daemon for insertion into the CSMDB. The size of that file can be controlled to create a set of test cases. I think @NickyDaB should have more info about how to create a test case from that.

pdlun92:
As described in #95 and #96 

Bucket is functional, so if this looks thorough enough to you @lasch I will merge this PR and close #96

### DIFF
--- a/csmtest/buckets/error_injection/messaging.sh
+++ b/csmtest/buckets/error_injection/messaging.sh
@@ -1,0 +1,82 @@
+#================================================================================
+#   
+#    buckets/error_injection/messaging.sh
+# 
+#  © Copyright IBM Corporation 2015-2018. All Rights Reserved
+#
+#    This program is licensed under the terms of the Eclipse Public License
+#    v1.0 as published by the Eclipse Foundation and available at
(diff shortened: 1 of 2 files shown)


pdlun92:
@lasch changed it to "partial" to avoid confusion

PR #134, opened 2018-06-27
bb:After processing all async requests, record the correct file seqnb…

dlherms-ibm:
…r/offset in the wrkqmgr for those request(s) just pushed to the HPWrkQ


sanomiya:
(used the wrong issue to commit my changes)

PR #156, opened 2018-07-12
Efix for burst buffer

tgooding:
Resolves several issues identified by the labs in the LSF->BB interface scripts.  Bumps efix level to 2 (--> cast 1.1.2)


tgooding:
Discussed with Rich and we'll create a separate branch for this particular efix.

Issue #108, opened 2018-06-12
Feature Request: set a string value to NULL


NickyDaB:
**Request:** Reset a field in the CSM Database back to NULL. (aka blank)

**Objective:** CSM update APIs should have the ability to set values to NULL in the CSM database. 

**Notes:**
- obviously not all CSM APIs should be able to set values in database to NULL
- only APIs with the tag of "update"?
- and even then, not all update APIs should let you set a value to NULL in the database.
- the important thing here is the feature, once the ability to implement the feature is abailable, then individual APIs can incorporate it, for specific attributes. 

**ToDo:**
- [x] test with a single API - node_attributes_update
- [x] test a single field - comment 
- [x] print the keyword on the back end logs (prove we got there)
- [x] update the database
- [x] pass a NULL value into a CSM API to set the value in the database to be NULL. 
- [x] support multiple fields on that API
- [x] support multiple APIs
- [x] improve re-usability of this new feature. 

**Feature branch associated with this issue:** https://github.com/NickyDaB/CAST/tree/feature-new/NULL_Database_fields_CSMAPIS

fpizzano:
Adding Py to this issue @watson6282.

NickyDaB:
as far as the `todo` item for `improve re-usability of this new feature.` we created a new issue for this, that will be worked on at a future release point.

NickyDaB:
@watson6282 @fpizzano  this feature is now in the development branch and will be available as part of the next CSM release

NickyDaB:
closing this issue.

PR #169, opened 2018-07-20
Feature new/null database fields csmapis

NickyDaB:
**Summary:**
- Adds in the ability to reset fields in the CSM Database to NULL via a csm API. 
- Reset a field to NULL with the special keyword "#CSM_NULL"
- designed for strings and text fields
- new function added in to do the comparing `CAST_stringTools_CSM_KEYWORD_Compare(input->comment, &keyword_compareCode)`
- this compares the string to all the CSM keywords we currently have
- it takes in the string, and a reference to an int, which the function uses to give the user some feedback.
- function is located in the `utilities/include/string_tools.h` file
- There is a RAS API that also resets a field back to NULL, but that wasn't a string; we use an ENUM in that case. 

**Reason behind Pull Request:**
Support for this feature was requested by the labs. Original issue can be found here: https://github.com/IBM/CAST/issues/108


utilities/src/string_tools.c
```diff
@@ -247,3 +247,79 @@ int CORAL_stringTools_nodeCount_xCATSyntax(char* myString, int* dataCount)
...
+*   1 keyword detected but no known match
+*   2 #CSM_NULL
+*
+* Returns: int 
+*   0 Success
+*   1 ERROR: Generic error. Default.
+*   2 ERROR: myString is NULL. Can not parse NULL string.
+*   3 ERROR: reached end of program without returning valid data
+*/
+int CAST_stringTools_CSM_KEYWORD_Compare(char* myString, int* compareCode)
```
mew2057: We should probably remove this with an enumerated type in the future. (Low Priority)

csmd/src/daemon/src/csmi_request_handler/CSMISwitchAttributesUpdate.cc
```diff
@@ -36,64 +36,113 @@ bool CSMISwitchAttributesUpdate::CreatePayload(
...
+	bool comment_NULL = false;
+
+	std::string stmt = "WITH updated AS ( UPDATE csm_switch SET ";
+	
+	//helper for keyword compare.
+	int keyword_returnCode = 0;
+	int keyword_compareCode = 0;
+	
+	//check to see if we have to do anything for comment. 
+	if(input->comment[0] != '\0')
```
mew2057: Is there a way to reduce the LOC for this, maybe make it more function driven? (Low priority)

NickyDaB: I thought about it. Almost made a macro, but I decided to go with this for future "edit-ability". This way if someone needs to come in and add specific functionality, then it is already all set up. And they can just insert the functionality code and not have to write logic code.

We could reduce the switch statement down, or marco-ise it, but then if someone wants to add in that extra functionality they would need to spend the extra time to expand it out. Based off of previous development cycles and requests for features on this project. I think this is the best move to mitigate future hassle. 

Here the switch statement already includes current possible outputs with comments describing each number and what they mean, so a new programmer on the project wouldn't have to explore code to look up the function either. 

But maybe there is a way to reduce lines without impact usability for the future programmer. I didn't think of any at the moment.

utilities/src/string_tools.c
NickyDaB: I agree. In the future we should have a file or list of all the `CSM_KEYWORDS` which are enums. I'll create an issue for it.

NickyDaB:
merging in

PR #174, opened 2018-07-21
Feature/bds access

lasch:
This fixes a critical problem with the BDS connection. Without this fix, if the connection fails, the bdsmgr component of csmd will not reconnect and will not even detect the failed connection properly in all cases.
This also updates the bds log msg in case the sending of data fails.


lasch:
In case anyone wants to test this in combination with the env-bds integration branch from @besawn, I've created a convenience branch in my fork where both parts are merged on top of the current master.
https://github.com/lasch/CAST/tree/feature/bds_integration

Issue #154, opened 2018-07-11
Python Analytics Use Case Conversion


mew2057:
The old Log Analysis use cases need to be converted to be used in the ELK stack. The following issue outlines the behavior and parameters for the existing use cases.

A sample use case may be found in my [branched copy of this repo](https://github.com/mew2057/CAST/blob/bds-usecases/csm_big_data/python/findJobKeys.py).

The original use cases may be found [here](https://github.com/IBM/CAST/tree/master/csm_big_data/Python/usecases).

## Use Case 1
**Script Name** : findJobTimeRange.py
- [x] Complete
**Description** : `A tool for finding when a job was running through the big data store.`
1. Queries `cast-allocation` for an `allocation id` or `primary job id`/ `secondary job id` combo.
2. Display the start and end time of the job

**Arguments**:
``` 
 -h, --help            show this help message and exit
  -a int, --allocationid int
                        The allocation ID of the job.
  -j int, --jobid int   The job ID of the job.
  -s int, --jobidsecondary int
                        The secondary job ID of the job (default : 0).
  -t hostname:port, --target hostname:port
                        An Elasticsearch server to be queried. This defaults
                        to the contents of environment variable
                        "CAST_ELASTIC".
  -H [host [host ...]], --hostnames [host [host ...]]
                        A list of hostnames to filter the results to
```

## Use Case 2
**Script Name** : findJobKeys.py
- [x] Complete
**Description** : `A tool for finding keywords during the run time of a job.`
1. Queries `cast-allocation` for an `allocation id` or `primary job id`/ `secondary job id` combo.
2. Use the allocation to filter logs on `hostname` and `time range` and search for any messages matching a list of `keywords` (`*` is wild).
3. Display keyword statistics.

**Arguments**:
``` 
 -h, --help            show this help message and exit
  -a int, --allocationid int
                        The allocation ID of the job.
  -j int, --jobid int   The job ID of the job.
  -s int, --jobidsecondary int
                        The secondary job ID of the job (default : 0).
  -t hostname:port, --target hostname:port
                        An Elasticsearch server to be queried. This defaults
                        to the contents of environment variable
                        "CAST_ELASTIC".
  -k [key [key ...]], --keywords [key [key ...]]
                        A list of keywords to search for in the Big Data Store
                        (default : *).
  -H [host [host ...]], --hostnames [host [host ...]]
                        A list of hostnames to filter the results to
```

## Use Case 3
**Script Name** : findJobMetrics.py
[...]

PR #208, opened 2018-08-08
Elastic Search Use Cases

etsai7:
# Adding Elastic Search Use Cases

## Purpose
The old Log Analysis use cases need to be converted to be used in the ELK stack. The following issue outlines the behavior and parameters for the existing use cases. 
- Issue: https://github.com/IBM/CAST/issues/154

## Approach
Using Python to write queries for Elastic Search. 

- [x] Use github checklists. When solved, check the box and explain the answer.
- [x] Assign both @pdlun92 and @williammorrison2 to review (So they can check regression)
- [x] Assign 1 developer to sanity check the code. (Working on automatically doing this via `./github/CODEOWNERS.md`)
- [x] Appropriatly label
- [ ] Assign a milestone. 

## Note
Use Case 3 still needs to query the data from extracted node data

### DIFF
--- a/csm_big_data/Use-Cases/ReadMe.md
+++ b/csm_big_data/Use-Cases/ReadMe.md
@@ -1,0 +1,107 @@
+## Use Case 1
+**Script Name**: findJobTimeRange.py
+
+**Description** : `A tool for finding when a job was running through the big data store.`
+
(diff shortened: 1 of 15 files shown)


mew2057:
I'm merging this in so we can continue work on the use cases. Thanks for your hard work @etsai7.

PR #228, opened 2018-08-24
UPDATES TO: Feature new/csm node find job

NickyDaB:
# UPDATES TO: Feature new/csm node find job

## Purpose
completes the additional functionality suggested for this new feature. 

## Approach
- adds basic requested functionality. any more requested would need to go through a user testing phase. 

- [ ] @pdlun92  please add test cases for all these new features

## How to Test

1. lots of new features. all need testing. too many to list. 
2. first pass can be testing all the different input options and cmd line flags.


csmd/src/daemon/src/csmi_request_handler/CSMINodeFindJob.cc
```diff
@@ -69,8 +69,8 @@ bool CSMINodeFindJob::CreatePayload(
 	int SQLNum_begin_time_search_end = 0;
 	int SQLNum_end_time_search_begin = 0;
 	int SQLNum_end_time_search_end = 0;
-	//int SQLNum_midpoint = 0;
-	//int SQLNum_midpoint_delta = 0;
+	int SQLNum_midpoint = 0;
```
mew2057: Didn't notice this on the last review, but there are a lot of variables here. This could make things more difficult to change in the future if needed. Not something that needs to be addressed now, but I'm just noting a [code smell](https://en.wikipedia.org/wiki/Code_smell) that contributes to the need for a better methodology of handling SQL across the board.

csmd/src/daemon/src/csmi_request_handler/CSMINodeFindJob.cc
```diff
@@ -286,11 +390,90 @@ bool CSMINodeFindJob::CreatePayload(
...
+								stmt.append(" ");
+							}
+						stmt.append(")");
+					}
+				stmt.append(
+				")");
+			}
+		stmt.append(") ");
+		stmt.append("ORDER BY ");
+		switch (input->order_by)
```
mew2057: Not something that needs to be fixed this run, but this should be swapped to either a function or static constant map. I think a function might be best, as it would allow us to drop this utility into any API and not have to worry about copy/paste errors.

csmi/src/inv/cmd/node_find_job.c
```diff
@@ -65,44 +65,62 @@ void help() {
...
 	puts("    csm_diag_run_query can have 2 optional filters.");
-	puts("    Argument      | Example value | Description  ");                                                 
-	puts("    --------------|---------------|--------------");
-	puts("    -l, --limit   | 10            | (INTEGER) SQL 'LIMIT' numeric value.");
-    puts("    -o, --offset  | 1             | (INTEGER) SQL 'OFFSET' numeric value.");
+	puts("    Argument       | Example value | Description  ");                                                 
+	puts("    ---------------|---------------|--------------");
+	puts("    -l, --limit    | 10            | (INTEGER) SQL 'LIMIT' numeric value.");
+    puts("    -o, --offset   | 1             | (INTEGER) SQL 'OFFSET' numeric value.");
+	puts("    -O, --order_by | a             | (CHAR) SQL 'ORDER BY' numeric value. Default Value: 'a'");
```
mew2057: As with the comment above, it might be a good idea to make this print out a function or macro in the future to standardize this feature.

csmd/src/daemon/src/csmi_request_handler/CSMINodeFindJob.cc
NickyDaB: We could expand this out in the future for all APIs, but at the moment I'm not sure if any other APIs would benefit. There was a fundamental issue with the pattern we construct SQL statements that didn't work in this API. 

Issue: I had to reuse an input variable (some of the search times) multiple times at different places in the SQL statement. In order to make sure that I was using the correct $number at the proper place in the SQL statement I had to save the position of all the input variables before hand, so that when I called 

```
// Build the parameterized list.
csm::db::DBReqContent *dbReq = new csm::db::DBReqContent(stmt, SQLparameterCount); 
```

at around line 485, that I added the variables in the correct order. 

But like I said before, I believe this is the only API that uses the same input variable multiple times at different places in the full SQL statement. 

And if needed, it could potentially make future changes easier, following a pattern of:
1.) get SQL positions of all input variables
2.) construct SQL statement
3.) add parameterized variables in the correct predetermined order

csmd/src/daemon/src/csmi_request_handler/CSMINodeFindJob.cc
NickyDaB: true. Yet at the moment, the order by seem to be unique to each API. as different APIs offer different content to order by. maybe after the 3rd or 5th order by, if I notice a pattern, I will expand it out into a function that builds it based off of the return record columns or a custom list.

csmi/src/inv/cmd/node_find_job.c
NickyDaB: noted for future design work.

csmi/src/inv/cmd/node_find_job.c
```diff
@@ -65,44 +65,62 @@ void help() {
 	puts("    -e, --end_time_search_begin   | \"2000-01-15 12:34:56.123456\" | (STRING) A time used to filter results of the SQL query and only include records with an end_time at or after (ie: '>=' ) this time.");
 	puts("    -E, --end_time_search_end     | \"2000-01-15 13:37:33.134317\" | (STRING) A time used to filter results of the SQL query and only include records with an end_time at or before (ie: '<=' ) this time.");
 	puts("    -m, --midpoint                | \"2000-01-15 12:00:00.000000\" | (STRING) A time used to filter results of the SQL query.");
```
pdlun92: Clean up this help comment

Issue #128, opened 2018-06-20
CSM Utility Daemon is unable to reconnect to CSM Master Daemon


besawn:
The problem was originally detected via a failure of an LSF daemon to communicate with the local CSM Utility daemon. Attempts to manually run /opt/ibm/csm/bin/csm_node_resources_query_all (or any other CSM command line program) on the utility node in question fail with the following message:
```
$ /opt/ibm/csm/bin/csm_node_resources_query_all

[csmapi][error] csmlib: Failed to connect to CSM_SSOCKET=/run/csmd.sock; retried 9 times.
csm_net_unix_Connect: No such file or directory
[csmapi][error] csm_net_unix_Connect() failed: /run/csmd.sock
[csmapi][error] /home/ppsbld/workspace/PUBLIC_CAST_V1.1.1_ppc64LE_RH7.5_ProdBuild/csmi/src/wm/cmd/node_resources_query_all.c-166:
[csmapi][error]   csm_init_lib rc= 2, Initialization failed. Success is required to be able to communicate between library and daemon. Are the daemons running?
```

The message above is considered a normal response when the local CSM daemon is unable to establish a connection with the CSM Master or Aggregator daemon. When there is a daemon-to-daemon communication failure, CSM intentionally stops listening for local client connections until the daemon-to-daemon communication is restored.

Through examination of the CSM Utility and Master daemon logs, the following behavior is observed:
- The Utility daemon is started ~10 minutes before the problem occurs.
- The Utility daemon is able to successfully connect to the Master daemon and process several message over several minutes.
- An unknown error occurs, causing the Utility daemon to detect the connection with the Master daemon is down via a failed heartbeat.
- The Utility daemon closes the connection to the Master daemon.
- The Utility daemon stops accepting local client requests, as designed. Local client API calls all fail with the message above.
- The Utiltity daemon then enters a retry loop, where it attempts to reconnect to the Master daemon, ~ every 50 seconds. However, it is never able to successfully reconnect, despite retrying for over 2 hours.
- Finally, a restart of just the Utility daemon allows it to successfully reconnect to the Master and normal operation is restored.

During each retry cycle, the following pattern is observed:
- The Utility daemon creates a new connection to the Master daemon and sends a version message.
- The Master daemon accepts the connection and processes the version message.
- Somewhere between the two daemons, the response from the version message from the Master to Utility daemon is lost, causing the Utility daemon to detect the connection as down and close it.

Open questions:
1.) What was the initial error that triggered this behavior?
[...]

PR #240, opened 2018-08-30
non-blocking connection review

lasch:
# non-blocking connection review

## Purpose
Reviewed the handling of non-blocking connect() for SSL and non-SSL connections between daemons. Potentially has an impact on issue #128 because there were some incorrect socket settings that got fixed.

Needs thorough testing especially with error cases where either the network or the peer daemon is not functioning properly. 

## How to Test
- regular function needs to be preserved
- in case a remote daemon is not responsive (stopped via signal, net cable pulled, firewall up), connection timeouts should be short to reduce the time the network thread waits for the timeout.
- especially important for compute daemons with 2 aggregators where one aggregator is unresponsive
- SSL connection becomes non-blocking with this PR which makes it useful to include SSL error cases too.

### DIFF
--- a/csmnet/src/CPP/endpoint_ptp.h
+++ b/csmnet/src/CPP/endpoint_ptp.h
@@ -101,13 +101,6 @@ protected:
     if( rc )
       throw csm::network::Exception("setsocket: SO_REUSEADDR");
 
-    // make the socket non-blocking for the time of the connect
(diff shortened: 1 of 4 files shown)


pdlun92:
Passes regression, so I think we can check off the "regular function needs to be preserved" box.  Will begin testing the core functionality of the fix in the error cases ASAP

PR #246, opened 2018-09-11
Recurring tasks

lasch:
# introducing interval trigger for recurring tasks to csm infrastructure

## Purpose
Enable to implement recurring tasks in a handler and configure the interval

## Approach
Added a new event source for special timer events that trigger a dedicated handler in each daemon.
The interval can be configured via config file.
If it's not configured, the feature is completely disabled.

- The dedicated handler currently only prints a log message when it gets triggered. It's subject to future work/decisions about what to do in that handler. There's one prepared (empty) handler class per daemon role.

## How to Test
- Run any daemon without config changes, nothing should change
- Configure the feature expect the log line to appear accordingly with the following config section, it should appear once every minute:
```
   csm {
       ...
        "recurring_tasks":
        {
            "soft_fail_recovery" :
            {
                "interval" : "00:01:00",
                "retry" : 3
            }
        }
      ...
```

## Screenshots
```
2018-09-05 14:24:23.000612       csmd::info     | INTERVAL_HDLR: INTERVAL: triggered handler to process
```

### DIFF
--- a/csmd/src/daemon/src/csm_event_source_set.cc
+++ b/csmd/src/daemon/src/csm_event_source_set.cc
@@ -42,6 +42,8 @@ EventSourceSet::EventSourceSet()
     MaxInterval = std::max(MaxInterval, (uint64_t)it);
   MaxInterval = std::max( MaxInterval, (uint64_t)DEFAULT_NETWORK_SRC_INTERVAL );
   MaxInterval = std::max( MaxInterval, (uint64_t)DEFAULT_TIMER_SRC_INTERVAL );
+  if( config->GetRole() == CSM_DAEMON_ROLE_MASTER )
+    MaxInterval = std::max( MaxInterval, (uint64_t)DEFAULT_DB_SRC_INTERVAL );
 
   RecurringTasks RT = config->GetRecurringTasks();
   if( RT.IsEnabled() )
@@ -54,6 +56,8 @@ EventSourceSet::EventSourceSet()
 
   mBucketScheduler->AddItem( NETWORK_SRC_ID, DEFAULT_NETWORK_SRC_INTERVAL, 0 );
   mBucketScheduler->AddItem( TIMER_SRC_ID, DEFAULT_TIMER_SRC_INTERVAL, 0 );
+  if( config->GetRole() == CSM_DAEMON_ROLE_MASTER )
+    mBucketScheduler->AddItem( DB_SRC_ID, DEFAULT_DB_SRC_INTERVAL, 0 );
 
   if( RT.IsEnabled() )
     mBucketScheduler->AddItem( INTERVAL_SRC_ID, RT.GetMinInterval(), 0 );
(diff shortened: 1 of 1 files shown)


csmd/src/daemon/src/csm_daemon_config.cc
```diff
@@ -1172,22 +1174,55 @@ void Configuration::CreateThreadPool()
...
+      boost::posix_time::time_duration interval_time = boost::posix_time::duration_from_string(ckey_str);
+      unsigned interval = interval_time.total_seconds();
+
+      ckey_str = GetValueInConfig( "csm.recurring_tasks.soft_fail_recovery.retry" );
+      unsigned retry = strtol( ckey_str.c_str(), nullptr, 10 );
+
+      if(( interval == 0 ) || ( retry == 0 ))
+      {
+        _Cron.Disable();
+        throw csm::daemon::Exception("Recurring Tasks configuration error. Interval/Retry invalid." );
```
besawn: Is there any way to disable a recurring task and leave the boilerplate for recurring tasks in the configuration file or does the section have to be completely removed from the configuration file to disable a recurring task?

When looking at the configuration file, I would have assumed the best way to disable a recurring task would be to set the interval to 0 or retry to 0 or both. Based on my understanding of this block of code, this will disable the function, but also cause the daemon to exit with a fatal exception.

lasch: Good point. Currently, the only way to disable is to remove the section (or rename it).
We can add an optional parameter like: csm.recurring_tasks.enable = true|false. To allow switching without removal.
I wouldn't recommend a way to disable it by setting interval/retry to 0 because the current entry for soft_fail_recovery might be just one out several. And then which one would you set to 0 in order to disable the feature?

besawn: Using an enable flag seems reasonable to me, but perhaps the scope should be at the level of the individual task so that soft_fail_recovery can be disabled but future_fundamental_feature can continue to run? 

csm.recurring_tasks.soft_fail_recovery.enable = true|false

Having an enable/disable option seems like a nice to have enhancement, but I think this pull request can be merged as-is with the new function added some time later.

lasch: ok. I think, we have to cover 2 different levels of "enable/disable".
The first level is the individual items like the soft_fail_recovery or any other future item.
The second level is the daemons event trigger and kick-off of the handler. We want the daemon to stay quiet (no events of this type) without a recurring task activated.

lasch:
@pdlun92 Yes. I noticed that one too. Trying to find the cause. Essentially it fails a check that tests whether all event sources had been scheduled at least once after one pass through the enabled sources. Thanks for the tests.

lasch:
@pdlun92 Updated code should no longer trigger the exception. At least, I can't reproduce in my env any more. 

For the records: The cause was that, on the master, the source for database events wasn't properly included in the scheduling. It just did not trigger a problem because until the introduction of the recurring event source, there was no active 'one-time source' at the master ('one-time sources' like the environmental data collection get scheduled only once per scheduling window).

pdlun92:
Looks good in all cases (all daemons with and without config section) after latest commit.  Approving

Issue #238, opened 2018-08-29
Python CSM API wrappers


NickyDaB:
**Summary:**
Python CSM API wrappers. Feature requested by labs and by @fpizzano 

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Issue Source:**
Reasoning for opening this issue; a description of why this issue was opened; a link to another thread that spawned the need for opening this issue.

**ToDo:**
- [ ] get original spec for this request and better fill in the above.

mew2057:
There's a fair amount of headway that's been made on this process using  
[boost python bindings](https://wiki.python.org/moin/boost.python/GettingStarted). The key is that this python binding is supposed to be a python binding of the C API.

PR #264, opened 2018-09-19
Research/csm python apis

NickyDaB:
# CSM APIs Python library Guide

## Purpose
- give our users more guidance on how to use CSM APIs through python. 
- addresses issue: https://github.com/IBM/CAST/issues/238
- addresses issue: https://github.com/NickyDaB/CAST/issues/14

## Approach
- This pull request gives our user base a short and direct guide on how to use the CSM API python library. 
- Adds in some more samples.

- [ ] In a future CSM release, we plan to make 1-1 matches of CSM API binaries with python scripts to further show examples of CSM APIs in Python. 

## How to Test

1. Is this guide straight forward?
1. Is it understandable?
1. Can you write a small script in python that uses the csm library?


csmi/python_samples/node_attributes_query.py
```diff
@@ -0,0 +1,64 @@
...
+#    v1.0 as published by the Eclipse Foundation and available at
+#    http://www.eclipse.org/legal/epl-v10.html
+#
+#    U.S. Government Users Restricted Rights:  Use, duplication or disclosure
+#    restricted by GSA ADP Schedule Contract with IBM Corp.
+#
+#================================================================================
+
+import sys
+sys.path.append('.')
```
mew2057: This isn't needed, this line was  leftover from a development test case.

csmi/python_samples/node_attributes_query.py
```diff
@@ -0,0 +1,64 @@
...
+
+#add the python library to the path 
+sys.path.append('/u/nbuonar/repos/CAST/work/csm/lib')
```
mew2057: In the future this directory should be added to the path by the rpm.

NickyDaB:
I'll work on implementing these changes when I can pull down the updated python directories. until then. my repo keeps wanting me to commit that conflicting cmake file.

NickyDaB:
@lasch 

> Why does the PR include the commits of several previous PRs that already had been merged to master? Is it because your PR would merge everything into the development branch?
Wouldn't it be better to separate the dev-master sync and your new updates?

yeah. that's what happened. I wanted some fixes that john put into master. and it grabbed everything. and then it kicked off some automation I had set up in development. (because this is still pull requesting into development) 

ill prob switch this pull request into master so i dont have to deal with all of this right now.

lasch:
Looking at the current locations of development and master branch, you might solve it by just forwarding the development branch to the current master (git merge --ff-only ....). I expect that to work because from what I see, the master is a direct descendant of the development location.

NickyDaB:
@lasch 
>Looking at the current locations of development and master branch, you might solve it by just forwarding the development branch to the current master (git merge --ff-only ....). I expect that to work because from what I see, the master is a direct descendant of the development location.

could try. but i'm not sure it would work because we made dev protected. and prevented merging directly into it without pull request code owner approval.

csmi/python_samples/guide.md
```diff
@@ -0,0 +1,206 @@
...
+
+When using the CSM Python library arrays must be `set` and `get`. 
+
+```Python
+nodes=["allie","node_01","bobby"]
+input.set_node_names(nodes)
+input.state = csm.csmi_node_state_t.CSM_NODE_IN_SERVICE
+```
+
+First we create an awway in Python. `nodes=["allie","node_01","bobby"]` Then we use the CSM Python library function `set_ARRAYNAME(array)` to set the `node_names` field of `input`. We do not need to set `node_names_count` like we do in C. the `set_` function will take care of that for you. Finally, we call `input.state = csm.csmi_node_state_t.CSM_NODE_IN_SERVICE` to set the state field of input to `IN_SERVICE`. This will tell CSM to set these 3 nodes to `IN_SERVICE`.
```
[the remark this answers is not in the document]
NickyDaB: addressed in: f76a3dd

csmi/python_samples/guide.md
```diff
@@ -0,0 +1,206 @@
...
+
+In the next line of code we call the csm API passing in the input we just populated. 
+
+```Python
+rc,handler,output = inv.node_attributes_update(input)
+```
+
+Our CSM library returns 3 values. 
+* A return code - Here defined as `rc`. This is the same as the return code found in the C version of the API.
+* A handler -
```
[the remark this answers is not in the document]
NickyDaB: addressed in: 
e5ab4f6

csmi/python_samples/node_attributes_query.py
NickyDaB: addressed in: b8d6217

csmi/python_samples/node_attributes_query.py
NickyDaB: addressed in: b8d6217 
- along with guide post that said you need it. also updated other sample files

csmi/python_samples/node_attributes_query.py
NickyDaB: noted. already talked to fernando about it. I have to make sure the libraries get copied to the right spot in an rpm. when i confirm that i will change this and the guide.

lasch:
> could try. but i'm not sure it would work because we made dev protected. and prevented merging directly into it without pull request code owner approval.

Agree, might be tricky to do in that case. However, maybe it makes sense to allow direct commits by one branch maintainer. Not my call though.

NickyDaB:
@lasch 

> Agree, might be tricky to do in that case. However, maybe it makes sense to allow direct commits by one branch maintainer. Not my call though.

We can set it up for admins can circumvent the protections. @tgooding is our only admin. We would have to nag him for when we need exceptions to the rule, which would be annoying for him i would assume. plus would limit to when we can do it to only when he is around. 

We could also make @fpizzano  an admin. That way we don't have to nag tom with CSM stuff.

tgooding:
This is probably an offline topic, but i'd like to understand how you did the following:
> some fixes that john put into master. and it grabbed everything

Whether you used cherry-pick, reverse pull request, or ???.  

Anyway, Rich Zollo (castbld) and Fernando (fpizzano) have admin powers.  But if you want to nag me to do a protected commit...  send a suitable offering of cookies to my cube.  :)

NickyDaB:
@tgooding 

> This is probably an offline topic, but i'd like to understand how you did the following:
>> some fixes that john put into master. and it grabbed everything

> Whether you used cherry-pick, reverse pull request, or ???.

I merged master into development.  I wasn't sure what I needed and what was dependent so I grabbed it all.

NickyDaB:
The tests now pass again. Good progress.

NickyDaB:
updated guide with new proper directory for libraries

PR #265, opened 2018-09-19
Removing old python directories

mew2057:
Removed old python directory to resolve windows collision.


NickyDaB:
Should we wait for @pdlun92  to run regression before we merge? Just to make sure the folder structure change doesn't mess with anything?

mew2057:
It's in the bds rpm which doesn't have a regression bucket ready and it looks like jenkins has already built so we should be fine.

Issue #249, opened 2018-09-13
Stale data on /etc/pam.d/csm/activelist from previous allocation


fpizzano:
Stale data on /etc/pam.d/csm/activelist from previous allocation.

```
# pdsh -w sierra[1210-1219,1720-1729,2470-2475,2976-2979] cat /etc/pam.d/csm/activelist | dshbak -c
----------------
sierra[1210-1215,1217-1219,1720-1728,2470-2471,2473,2976-2979]
----------------
nvidia;132489
----------------
sierra[1216,2474]
----------------
nvidia;132401
nvidia;132489
----------------
sierra[1729,2475]
----------------
nvidia;132404
nvidia;132489
----------------
sierra2472
----------------
nvidia;132406
nvidia;132489
```

fpizzano:
Adding @adambertsch @watson6282

mew2057:
@fpizzano and I discussed this last week, resolution is going to be multiphase:

- [ ]  On non-shared allocations blank this file.
- [ ]  On non-shared allocations delete any lingering cgroups.
- [ ] Verify that missing cgroups do not place the user into the system cgroup.

adambertsch:
The specific issue on Sierra was that there were two entries in activelist, one valid and one invalid, with the same user name, but different allocation IDs.  The invalid entry was masking the valid entry, which is a case that could be handled better, and then in turn the user was ending up in system cgroup.  It seems like there may be at least 2 fixes to make just in the code to parse/act on activelist.

mew2057:
I have a possible fix ready. I tested in my environment and it seems to work, need a regression from @pdlun92 to verify. We should try adding a test case where a bad allocation is in the activelist then check the login info afterwards to the test bucket in the future.

PR #268, opened 2018-09-20
PAM fixes

mew2057:
Added fixes to prevent the failures described in Issue #249. The following actions are now performed:

1. If the pam module fails to place the user in the CGroup, the search of the activelist continues.
2. When an allocation is created in the non shared state two new actions are performed:
  * The activelist is wiped.
  * The existing allocation cgroups are removed.
3. When an allocation is created with zero isolated cores the system cgroup is removed.

Additionally, the RPM has been modified to add the account to the ssh pam config and the documentation has been updated to reflect the change.


mew2057:
Total accident thanks for the catch!

mew2057:
@pdlun92 can you rerun regression, had to disable the csm_system removal.

pdlun92:
@mew2057 Re-ran regression, still looks good

PR #270, opened 2018-09-20
Feature improvement/ufm inventory collection master

NickyDaB:
# Add default values to CSM-UFM inventory collection

## Purpose
@fpizzano requested default values be added to the CSM UFM inventory collection program. No formal issue was created, so I can not tag it here. 

## Approach
- This is a simple addition that sets the values of the parameters before the program runs. If the user inputs parameters, then these values will override the defaults. 

## To Do:
- [x] @lasch please be the developer reviewer
- [x] @pdlun92 please run FVT regression
- [x] @williammorrison2  please run P8 regression 

## How to Test

1. Run the program without passing in parameters. Does it set values to default?
2. Run the program providing parameters. Does it override default values?
3. Run the program without passing in parameters where default location doesn't exist. Does it print error? 
4. Other tests that @pdlun92 would like to add. 

## Screenshots

Example of running program when providing a flag for the config file, but not providing a flag for the type of collection:
```
[root@c650f03p41 sbin]# ./standalone_ib_and_switch_collection -c /u/nbuonar/csm_master.cfg
Loading the inputs
full path to the master config file: /u/nbuonar/csm_master.cfg
Type of collection: 3


type of collection 3
ib flag 1
ib error full path: /var/log/ibm/csm/inv/bad_ib_cable_records.txt
reading input file

WARNING: 12 IB cables found with 'NA' serial numbers and have been removed from CSM inventory collection data.
These records copied into 'bad_ib_cable_records.txt' located at '/var/log/ibm/csm/inv/' 


UFM reported 95 IB records.

switch flag 1
output file name: output_file.txt
input file name: output_file.txt
reading input file...
totalIBRecords 83
totalSwitchRecords 6
---
# ib inventory successful
...
# total ib inventory collected: 83
# new ib records inserted into database: 83
# old ib records updated in database: 0
---
# switch inventory successful
...
---
# switch module inventory successful
...

```

Example of running program with default values when `/etc` folder doesn't exist.
```
[root@c650f03p41 sbin]# ./standalone_ib_and_switch_collection
Loading the inputs
full path to the master config file: /etc/ibm/csm/csm_master.cfg
Type of collection: 3

The master config file is not open
Did you provide the full correct path to the master config file?
Return without executing data collection
[root@c650f03p41 sbin]#
```


NickyDaB:
@lasch . Thanks. yeah. let's make it pretty.

NickyDaB:
@pdlun92 

>  The only quality of life change I would suggest would be a default location for the output file (that isn't just the current working directory). However that can certainly be addressed in a later pull request if it needs to be addressed at all.

I'll get it in by the end of the week.

PR #277, opened 2018-09-24
Fixed python library build

mew2057:
Python shared objects were not being built properly, new fix allows them to be built and places them in the api RPM.

### DIFF
--- a/csmd/src/daemon/src/csmi_request_handler/helpers/Agent.cc
+++ b/csmd/src/daemon/src/csmi_request_handler/helpers/Agent.cc
@@ -228,9 +228,16 @@ int ForkAndExecAllocationCGroup(char * const argv[], uint64_t allocation_id, uid
             try
             {
                 csm::daemon::helper::CGroup cgroup = csm::daemon::helper::CGroup( allocation_id );
-
+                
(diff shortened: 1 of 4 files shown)


csmd/src/daemon/src/csmi_request_handler/helpers/Agent.cc
```diff
@@ -228,9 +228,16 @@ int ForkAndExecAllocationCGroup(char * const argv[], uint64_t allocation_id, uid
             try
             {
                 csm::daemon::helper::CGroup cgroup = csm::daemon::helper::CGroup( allocation_id );
-
+                
+                LOG(csmapi, trace) << "Waiting on pid #" << getpid() << " migration allocation: " << allocation_id;
                 if(cgroup.WaitPidMigration(getpid()))
-                    _Exit(execv(*argv, argv));
```
NickyDaB: Was this accidentally committed? I don't see the obvious connection to the python library fix.

NickyDaB: not just this line. but the whole helpers agent file

Issue #248, opened 2018-09-13
CSM Aggregator might get stuck when trying to connect to BDS


lasch:
The CSM daemon (currently only aggregator) tries to connect to BDS if configured. If the BDS is not responding or there's network problems causing the connect() to wait, there's potential for the csmd to get stuck as well.

This is most critical during daemon startup, because the daemon tries to connect to BDS at the time the BDS-Mgr component of csmd is created. In this case, the startup of the daemon might get delayed until the connect() times out.
Later, the BDS-Mgr will retry the connect() in a separate thread, so it has a low impact on the operation of the daemon itself.  However, it might have an impact on the shutdown. If the BDS-Mgr thread waits for the connect() to return while the daemon is shut down, the csmd process might be unable to terminate and needs to be killed manually. This is especially critical when a daemon is restarted because the new csmd will likely fail to start properly if the old daemon process is still out there.

**To Reproduce**
1. Start an aggregator with configured BDS and have the BDS either stopped or suspended
2. Try to run a CSM API on the aggregator; or try to start a compute daemon that connects to this aggregator

**Expected behavior**
 - Step 2 would be unable to connect because the aggregator is not in running mode yet
 - with regular shutdown, the aggregator process might not terminate
I've seen in only occasionally happening while running no BDS process at all. So it might not reproduce 100% of the time.
On

**Environment:**
 - so far only on x86, i.e. my laptop, haven't tried anywhere else yet
 - Version: current head of master (should be this way since at least 1.2.0)

One cure will be to use a non-blocking connect() and set a low timeout to reduce the delay to only a few seconds or anything acceptable.

lasch:
Non blocking connection introduced to fix this. Closing.

PR #279, opened 2018-09-25
non-blocking BDS connect

lasch:
Non-blocking connection to BDS to solve issue #248 

## Approach
same mechanism as for daemon-to-daemon connections via non-blocking socket and check for activity via select with short timeout to avoid having to wait the default network timeout in worst case.

- So far, I haven't bothered to change the whole communication with BDS into non-blocking. So there's still some potential for send() operations to get stuck until timeout.

## How to Test
I have not been able to reliably reproduce the stuck connection.
Since we've never seen this in the wild, testing should focus on 'existing functionality preserved' for now. I'll update issue/PR/test team if I stumble over the problem again

## note:
the diff of endpoint_ptp.h shows an unrelated function being moved when actually CheckConnectActivity() has been moved to public and made static inline.

### DIFF
--- a/csmd/src/daemon/src/csm_bds_manager.cc
+++ b/csmd/src/daemon/src/csm_bds_manager.cc
@@ -236,6 +236,8 @@ csm::daemon::EventManagerBDS::SendData( const std::string data )
       remain -= rc;
       done += rc;
     }
+    if(( rc == -1 ) && ( errno == EINTR ))
+      rc = 0;
   }
 
   return ((rc >= 0) && ( remain == 0 ));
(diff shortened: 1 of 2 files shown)


pdlun92:
@lasch I got a chance to test this out a bit.  I tried to accomplish this by putting logstash and the aggregator daemon through various start/stop cycles.  I observed the aggregator logs on a build from master branch, and then on a build from master branch with these changes merged in.  The behaviors identical to me.  In particular, I was looking for `BDSMGR: Connected` messages on the first successful bucket send to BDS and for `BDSMGR: Connection to BDS failed: rc=0:0` on unsuccessful sends.  If you think that is sufficient, I will approve these changes and we can merge

PS: It passes regression as well

lasch:
Not sure about the second string you look for. The search should probably be for ```BDSMGR: Connection to BDS failed...```, because the ```rc=0:0``` could be changing to other values.
Anyway, I think, that's good enough for what we look for right now.

PR #280, opened 2018-09-25
add debug/test env-data source

lasch:
Adding a debug or test source for environmental data collection and BDS testing.

When running on a system that's not P9 and/or has no DCGM active, there will be no environmental data collection visible to the user or the aggregator or to BDS. This PR adds a DEBUG source that sends some fixed data (+time stamp and host name) an an env-data dummy.
The sole purpose is for testing if the env data from the compute node arrives in BDS at the configured interval for testbeds that have no hardware or firmware support for our existing data collection.

When configuring:
```
        "data_collection" :
        {
            "buckets":
                [
                    {
                        "execution_interval":"00:00:10",
                        "item_list" : ["debug"]
                    }
                ]
         }
```

This is an example test string that would be sent to BDS every 10s:
```
{"type":"csm-test-env","source":"hostname","timestamp":"2018-09-25 15:07:12.367419","data":{"debug":"Fixed generated debug\/test data"}}
```

This is obviously a low priority PR. I'm not assigning a milestone to it.
@fpizzano should decide whether we want that added or not.

### DIFF
--- a/csmd/src/daemon/include/bucket_item_type_definitions.h
+++ b/csmd/src/daemon/include/bucket_item_type_definitions.h
@@ -13,6 +13,7 @@
 
 ================================================================================*/
 // USE UPPER CASE for each bucket item
+bucket_item(DEBUG)
 bucket_item(CPU)
 bucket_item(GPU)
 bucket_item(NETWORK)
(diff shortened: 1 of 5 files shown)


lasch:
@pdlun92 Given that it is a kind of 'hidden' feature, do we need regression runs on this before merging?
I guess, it would make sense to check if nothing else broke?

PR #283, opened 2018-09-25
csm: add switches to enable/disable recurring tasks in config

lasch:
Introduce switches to enable or disable recurring tasks separately or entirely in config

The defaults currently set: false for everything because the corresponding handler only is a placeholder for later development.

## How to Test
```
        "recurring_tasks":
        {
            "enabled" : false,
            "soft_fail_recovery" :
            {
                "enabled" : false,
                "interval" : "00:01:00",
                "retry" : 3
            }
        }
```
* With everything set to false, the log should print info-level msg that shows the feature disabled.
* same should happen if you enable just the soft-fail-recovery section
* If recurring.tasks.enabled = true, the log should print info-level msg that indicates the feature disabled because there are no active tasks
* if both is set true, you should see info-level msg showing the settings for soft-fail-recovery (and the values should match the config; also, every minute you should see output from the placeholder handler
```
2018-09-25 17:51:40.567310       csmd::info     | INTERVAL_HDLR: INTERVAL: triggered handler to process
```

### DIFF
--- a/csmconf/csm_master.cfg
+++ b/csmconf/csm_master.cfg
@@ -96,8 +96,10 @@
 
         "recurring_tasks":
         {
+            "enabled" : false,
             "soft_fail_recovery" :
             {
+                "enabled" : false,
                 "interval" : "00:01:00",
                 "retry" : 3
             }
--- a/csmd/src/daemon/include/csm_recurring_tasks.h
+++ b/csmd/src/daemon/include/csm_recurring_tasks.h
@@ -28,6 +28,7 @@ typedef struct
 {
   unsigned _Interval;
   unsigned _Retry;
+  bool _Enabled;
 } SoftFailRecovery_t;
 
 class RecurringTasks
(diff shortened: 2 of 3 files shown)


pdlun92:
For future reference (especially for adding to regression) here is a "truth table" for the log outputs:

1. recurring_tasks: *false* soft_fail_recovery: *false*
```
Configuration: Recurring tasks disabled or not defined in config. Disabling feature.
```
2. recurring_tasks: *false* soft_fail_recovery: *true*
```
Configuration: Recurring tasks disabled or not defined in config. Disabling feature.
```
3. recurring_tasks: *true* soft_fail_recovery: *false*
```
Configuration: Recurring task: Soft-failure recovery is disabled. Disabling recurring tasks entirely.
```
4. recurring_tasks: *true* soft_fail_recovery: *true*
```
INTERVAL_HDLR: INTERVAL: triggered handler to process
```

NickyDaB:
I'm sorry if I am out of the loop but what are "recurring tasks"?

lasch:
> I'm sorry if I am out of the loop but what are "recurring tasks"?

recurring tasks is the preparation of a feature that allows to trigger certain activities in a cron-like fashion with a configurable interval. the first candidate is a handler that would try to recover nodes from soft-failure status.

Issue #307, opened 2018-10-05
hcdiag: timeout does not kill running processes


Obihoernchen:
**Describe the bug**
hcdiag uses xdsh to execute commands remotely. Unfortunately the xdsh timeout argument (-t) doesn't work as you would expected. It does not kill the remote command.

A command looks like this:
```
└─python /opt/ibm/csm/hcdiag/bin/hcdiag_run.py --bucket burn_in --target taurusml[2-10]
    ├─perl /opt/xcat/bin/xdsh taurusml[2-10] -s -z -t 800 cd /opt/ibm/csm/hcdiag/tests/hxefpu64; ./hxefpu64.sh 10 1
```

Timeout is set to 800s but if the test does not finish in 800s it will keep running on the node.
_Note: I tested this without CSM. I couldn't test with CSM_

**To Reproduce**
Create a simple "sleep" test `/opt/ibm/csm/hcdiag/tests/test-sleep/test-sleep.sh`:
```bash
if [ -n "$HCDIAG_LOGDIR" ]; then
   [ ! -d "$HCDIAG_LOGDIR" ] && echo "Invalid directory. Exiting" && exit 1
   THIS_LOG=$HCDIAG_LOGDIR/`hostname -s`-`date +%Y-%m-%d-%H_%M_%S`.output
   echo $THIS_LOG
   exec 2>$THIS_LOG 1>&2
fi

model=$(grep model /proc/cpuinfo|cut -d ':' -f2)
readonly me=${0##*/}

echo -e "Running $me on $(hostname -s), machine type $model.\n"          
echo "from $me: hello world!"
# sleep for 1000s to demonstrate a timeout
sleep 1000
rc=$?
#rc=10
if [[ $rc -eq 0 ]]; then
   echo "$me test PASS, rc= $rc"
else
   echo "$me test FAIL, rc= $rc"
fi

exit $rc
```

Configure it in `test.properties` like this with 1s timeout:
```
[tests.test-sleep]
description = simple sleep test
executable  = /opt/ibm/csm/hcdiag/tests/test-sleep/test-sleep.sh
timeout     = 1
```

_Note: I don't have CSM, so I don't know if this issue exists with CSM, too._

Execute the test like this: 
```
/opt/ibm/csm/hcdiag/bin/hcdiag_run.py --nocsm --test test-sleep --target taurusml11
```

Go to node and you should see the process still running even so it's already timed out:
```
[root@taurusml11 ~]# ps -ef | grep sleep
[...]

PR #321, opened 2018-10-10
Issue #307: Added a timeout to the bb_cmd agent code,

mew2057:
Keys off of th timeout time in the configuration file. Issues a SIGKILL to the child process if the timer is exceeded.


tgooding:
Why the change from waitpid() to wait()?  Seems like thats going to cause problems in a multithreaded csmd and other processes coming and going.  wait() just knows that some child finished (e.g., if i'm reading right, an epilog script finishing 5 milliseconds after bbcmd was launched would kill the bbcmd very prematurely).

mew2057:
That's an accidental wait, meant to return it to waitpid, thanks for the catch.

PR #341, opened 2018-10-17
Environment JSON fixes

mew2057:
Updates to the regular expression to capture negative numbers and reject numbers with leading zeroes.


mew2057:
You can try it on https://regexr.com/ and it will explain the two cases.
``` regex
:\s*\"(([-]{0,1}(0)|([1-9]+[0-9]*))(\.[0-9]*)?)\" 
```

The key is this part: 
``` regex
((0)|([1-9]+[0-9]*))
```
Basically this says 0 *or* a whole number starting with 1 through 9. This excludes the `0123` case, but preserves the `0` or `0.1` cases.

besawn:
The explanation above makes sense to me. I had tried it out on a different regex testing site, but clearly not all regex testing sites are equal.

PR #343, opened 2018-10-17
Explicitly mark configuration file as readable by group/other

tgooding:
bb.cfg should be marked 0644, but its possible the bbactivate caller's umask may not be 0022.


tgooding:
Verified

PR #344, opened 2018-10-18
BDS msg caching at aggregator

lasch:
This pull request will add a caching feature to the aggregator.
If the aggregator has lost the connection to the BDS, it will cache any BDS messages (e.g. environmental data from compute nodes) for a configurable amount of time.

This will help to prevent BDS data to get dropped due to temporary failures or restarts of the BDS service.

The expiration time of cached messages can be configured in the config file via a new option:

```
        "bds" :
        {
                "host" : "__LOGSTASH__",
                "port" : 10522,
                "reconnect_interval_max" : 5,
                "data_cache_expiration" : 50
        }
```
* If the ```data_cache_expiration``` parameter is omitted, there will be a default expiration time that is 5x of the configured (or default) reconnect interval.
* If expiration is set to 0, then caching is disabled and data gets dropped while the BDS is down
* If the expiration is configured smaller than the reconnect interval, there will be a non-critical warning about potential data loss
* If the expiration is set to an interval longer than 1000s, there will be a non-critical warning about memory usage.

##How to test
- general test approach would probably be to stop or disconnect the BDS for a time that is shorter than the cache expiration and when BDS is brought back there should be no data missing.
- in case of a longer outage of the BDS, only data that has not expired in the cache would show up

1) run existing config without any configured expiration the time limit should be 5x of the reconnect interval
2) configure expiration of 0 and expect all data to get dropped while BDS is down
3) configure an expiration and expect BDS data to get cached and sent to BDS for that configured time span

Note: the expiration time isn't enforced down to the millisecond. So you might see cases where messages get sent to BDS even if they expired shortly before the reconnect. I intentionally didn't aim for high precision timing of the expiration because I think it's not necessary. If anyone thinks it should be enforced precisely, please let me know.

### DIFF
--- a/csmconf/csm_aggregator.cfg
+++ b/csmconf/csm_aggregator.cfg
@@ -58,7 +58,7 @@
                 "host" : "__LOGSTASH__",
                 "port" : 10522,
                 "reconnect_interval_max" : 5,
-                "data_cache_expiration" : 50
+                "data_cache_expiration" : 600
         }
     }
 }
--- a/csmconf/csm_aggregator.cfg
+++ b/csmconf/csm_aggregator.cfg
@@ -57,7 +57,8 @@
         {
                 "host" : "__LOGSTASH__",
                 "port" : 10522,
-                "reconnect_interval_max" : 5
+                "reconnect_interval_max" : 5,
+                "data_cache_expiration" : 50
         }
     }
 }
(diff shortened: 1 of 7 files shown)


lasch:
Updated the defaults and logging according to our discussion. Ready for review now.

lasch:
Just found one more issue that I should fix before merging.
We agreed to set the default cache expiration to 1000. However, if someone configures the reconnect interval but no cache expiration, then the default cache expiration will be 5x of that reconnect interval. I'll fix this later today.

lasch:
problem fixed. Defaults to 10min when no value provided in config file.

PR #350, opened 2018-10-18
Logstash and Kibana plugin RPMs

mew2057:
Both the kibana and logstash plugins are now `ibm-csm-bds-kibana-1.3.0-*.noarch.rpm` and `ibm-csm-bds-logstash-1.3.0-*.noarch.rpm`

A fix for `zimonCollector.py` was also added.


mew2057:
Merging to master so we can see the impact on the rpm build.

PR #379, opened 2018-10-30
Compare to `None` using identity `is` operator

janisozaur:
This is a trivial change that replaces `==` operator with `is` operator, following PEP 8 guideline:

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.

https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations


fpizzano:
@janisozaur  Thank you for your interest in this project and creating this PR.

To officially contribute to the project, we do need an CLA signed, if you are covered under a CCLA, please provide that information to the email addresses listed here: https://github.com/IBM/CAST/blob/master/CONTRIBUTING.md, we look forward to any contributions you can make in the future.

PR #390, opened 2018-11-07
csm: using steady clock

lasch:
This PR replaces csmd usage of system_clock with steady_clock to reduce the chances for hickups when timezone changes between daylight and standart time or the clock gets adjusted for other reasons.

Especially covering any adjustments where the new time is in the past. This would cause any timers to take longer to trigger causing delays in heartbeats (e.g. for connection error detection) and other timers that got set to observe API timeouts and more.

There are a few instances of system time left in the CSM API section that would have to be reviewed by corresponding code owners (probably @NickyDaB or @mew2057). As far as I can tell, the changes in this PR have no impact on the creation of any user-visible or BDS-inserted time stamps. @besawn might want to check that statement from the RAS and Inventory perspective.

Note: a forward adjustment of system time might still cause issues in the sense of timers that expire too early. However, I can only think of issues that would cause API timeouts. The connection heartbeats require 3 timer intervals to expire before shutting down connections and a single time adjustment would only affect one interval.

### DIFF
--- a/csmd/src/daemon/include/csm_bds_manager.h
+++ b/csmd/src/daemon/include/csm_bds_manager.h
@@ -113,7 +113,7 @@ public:
   bool AddToCache( const std::string bsd_msg );
 
 private:
-  inline void UpdateCurrentTime() { _CurrentTime = std::chrono::system_clock::now(); }
+  inline void UpdateCurrentTime() { _CurrentTime = std::chrono::steady_clock::now(); }
   inline BDSTimeType GetTimestamp() { return _CurrentTime; };
   void CheckCachedData();
 };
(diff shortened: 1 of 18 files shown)


pdlun92:
Passes regression.  Will work on some unit testing once I get through some more pull requests

NickyDaB:
> There are a few instances of system time left in the CSM API section that would have to be reviewed by corresponding code owners (probably @NickyDaB or @mew2057). 

Looks like the `infrastructure_check` handler inside of `csmi_request_handler` was edited. thats why i was added in as a code owner. (because i have a trigger on the folder) So I don't think any other APIs touched this pull request. Approving.

lasch:
nice. Thanks @pdlun92

PR #401, opened 2018-11-08
csmdb_update: fn_csm_switch_inventory_history_dump now handles NULL f…

williammorrison2:
…ields without creating duplicate records in the history table.

# Pull Request Title
bills_fn_csm_switch_inventory_history_dump_update

## Purpose
Updated the function to handle NULL fields without creating duplicate records in the history table.

- [ ] @pdlun92 to review (check regression)
- [ ] @NickyDaB to review with basic API testing + code review.

## How to Test
INSERT a record or records with potential 'NULL' field values.
THEN make sure there are non duplicated records in the history table when updating (Running inventory collection - If no fields have changed).


NickyDaB:
I tested it while @williammorrison2  was testing. It looks to me like the issue @fpizzano  had has now been resolved.

PR #402, opened 2018-11-08
a few small fixes

lasch:
@tgooding found a few small things should be cleaned up.

I combined them in a single PR to reduce the noise from git notifications.


pdlun92:
Passes regression

pdlun92:
Still passes regression. Not surprising but FYI

PR #409, opened 2018-11-14
Adding Compute Nodes to Error Output

mew2057:
Implemented a new mechanism to report errors from the compute daemon to the front end. 
Header Changes:
* Fixed a typo in the error codes (1.4.0 content)
  *`CSMERR_ALLOC_UNAVIL_NODES` -> `CSMERR_ALLOC_UNAVAIL_NODES`
* Converted the error handling to use the automated mechanism for serialization/deserialization.
* Added four new functions for processing errors `csmi/include/csm_api_common.h`:
 * `uint32_t csm_api_object_node_error_count_get(csm_api_object *csm_obj)`
 * `int csm_api_object_node_error_code_get(csm_api_object *csm_obj, uint32_t index)`
 * `const char* csm_api_object_node_error_source_get(csm_api_object *csm_obj, uint32_t index)`
 * `const char* csm_api_object_node_error_msg_get(csm_api_object *csm_obj, uint32_t index)`
* Added an internal mechanism for printing a list of node errors for command line APIs (csm_print_node_errors).

**NOTE** None of the external features have been converted to python at this time.


mew2057:
@lasch can you check out 377975ea7c6c8d1c1177832957c6193cf6b95a4d. I implemented a fix where I added a second print of the ctx object, because if two ctx prints aren't present (regardless of loggging level) we get a segmentation fault in this code block. 

It looks like the ctx object loses its reference after extracting the timer event content. I'm not sure why this happens, but it almost smells like a race condition or timing issue. I checked and it looks like there are no calls to the destructor of the object so I think it's a problem with the shared pointer.

I want to fix this for real, so the total PR is not up for review yet, I just wanted to expose the code for people to take a look.

mew2057:
So I'm trying it again and now it's not reproducing.

mew2057:
I think there was some weird memory collision happening in the cast of the event to a timer event, I've tried a new strategy for casting in `HandleTimeout`, hopefully that catches the issue. @pdlun92  should run regression with timeouts to verify.

mew2057:
I think this might introduce a memory leak, looking into it now.

PR #413, opened 2018-11-16
Jsm new params

mew2057:
Added hostnames to parameters to the jsm call.

This will look something like `--num_hosts 4 --hosts a b c d` to the jsm executable.


awlauria:
One minor change that we need for jsm is the launch node be at the head of the list of hosts passed in as arguments.

NickyDaB:
> One minor change that we need for jsm is the launch node be at the head of the list of hosts passed in as arguments.

Maybe have the launch node be a separate argument?

awlauria:
I'm fine with either adding the launch node as a separate argument or adding it to the head of the list.

It would make it more consistent with what is in the `csm_allocation_query_output_t` struct since it stores the launch node in a separate field.

mew2057:
I'm adding a `--launch_node` field to the call, it will have "BAD_NODE" in the unlikely event of no launch node being unspecified. I'm almost done with the code now.

mew2057:
The incoming arguments will look something like this: 

`--launch_node sierra --num_hosts 1 --hosts "c650f03p39-vm09 "`

@pdlun92 we might want to run regression again, pending whether @awlauria needs any additional changes.

csmd/src/daemon/src/csmi_request_handler/helpers/Agent.h
```diff
@@ -142,10 +142,43 @@ inline int ExecuteSFRecovery( char ** output, int timeout)
...
+    uint32_t num_nodes, char** compute_nodes, char* launch_node)
 {
+    // Build the nodes string.
+    std::string hosts = "\"";
+
+    for (size_t j=0; j < num_nodes; ++j)
+    {
+        hosts.append(compute_nodes[j]).append(" ");
+    }
+    hosts.append("\"");
```
awlauria: Can we do something like:

`hosts[hosts.length() - 1] = 0`

to get rid of that last " "?

awlauria:
Other than that I think this format looks good to me.

Thanks.

csmd/src/daemon/src/csmi_request_handler/helpers/Agent.h
```diff
@@ -146,13 +146,13 @@ inline int ExecuteJSRUN( char* jsm_path, int64_t allocation_id, uid_t user_id, c
...
     // Build the nodes string.
-    std::string hosts = "\"";
+    std::string hosts = "";
 
     for (size_t j=0; j < num_nodes; ++j)
     {
         hosts.append(compute_nodes[j]).append(" ");
     }
-    hosts.back() = '\"';
+    //hosts.back() = '\"';
```
awlauria: Can you replace this with:

`hosts.back() = 0`

To get rid of the space again. And I think we're golden

PR #432, opened 2018-12-04
Soft Failure Bug Fixes

mew2057:
Fixes for some of the Soft Failure Bugs uncovered in test:
1. Transitioned to use RAS events to trigger hard failure.
2. Enhanced debug lines in error output (fixed commas, non errors don't present error messages).
3. Enabled Soft Failure to Hard Failure through RAS events.
4. Added mechanism to push events from multicast customization.

Test cases that need to be made:
1. Call soft failure recovery with no nodes in soft failure (`CSMI_NO_RESULTS`).
2. Call soft failure recovery with one node in soft failure (`CSMI_SUCCESS`).
3. Call soft failure recovery with multiple nodes in soft failure (`CSMI_SUCCESS`).
4. Call soft failure recovery with one node in soft failure in an allocation (`CSMI_NO_RESULTS`).
5. Call soft failure recovery with one node in soft failure in an allocation and one in soft failure without an allocation (`CSMI_SUCCESS`).
6. Call soft failure recovery with a node with a shut down daemon (`CSMERR_TIMEOUT`).

@besawn I added some logic to enable the `Soft Failure` -> `Hard Failure`.

### DIFF
--- a/csmd/src/daemon/src/csmi_request_handler/CSMIRasEventCreate.cc
+++ b/csmd/src/daemon/src/csmi_request_handler/CSMIRasEventCreate.cc
@@ -421,7 +421,13 @@ csm::db::DBReqContent CSMIRasEventCreate::getRasCreateDbReq(RasEvent &rasEvent)
         sql_stmt << "$" << ++param << "::compute_node_states";     // SQL param: set_state
         sql_stmt <<  " where node_name=";
         sql_stmt << "$" << ++param << "::text";                    // SQL param: location_name
-        sql_stmt <<  " AND state IN ('" << CSM_NODE_STATE_DISCOVERED << "','" << CSM_NODE_STATE_IN_SERVICE << "') ) ";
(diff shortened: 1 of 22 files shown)


besawn:
@mew2057 I saw that changes would be required to allow the SOFT_FAILURE to HARD_FAILURE transition yesterday. I discussed with @fpizzano, and the plan as of yesterday was to think over the situation and then discuss amongst the three of us. 

I am not concerned with allowing the transition for the case where the soft failure recovery handler is creating the RAS event. I was more concerned about what other transitions will be allowed by this change. For example, without this change, once a node goes into SOFT_FAILURE, it will stick there until the soft failure recovery handler moves the node into either HARD_FAILURE or IN_SERVICE. With the change contained in this pull request, the soft failure recovery handler could be trying to recover a node, and a different RAS event will then put the node into HARD_FAILURE. This is the specific race condition I wanted to dig into in more detail.

Let's discuss tomorrow.

mew2057:
We could patch the race condition using filtering on the `SOFT_FAILURE` -> `IN_SERVICE` and give a precedence to `HARD_FAILURE`. In that case `HARD_FAILURE` would never be recoverable through automation.

PR #434, opened 2018-12-05
candidate fix for csm client-crash issue

lasch:
This introduces a candidate fix for a potential race condition.

The race condition can happen when the client is communicating with the CSM daemon and that daemon is forced into disconnected mode. (e.g. aggregators of a compute node become unreachable, or master node goes down so utility and aggregator become disconnected).

The approach is a mutex around send/recv operations and the socket shutdown/cleanup so that a socket that's in active recv/send can no longer run into wiped/freed data structures that got cleaned by the shutdown sequence.

I have not been able to reproduce that crash with or without this fix yet. It seems that this crash is a rare event since the unprotected code sections existed for quite a while.

### DIFF
--- a/csmnet/src/C/csm_network_internal_api_c.h
+++ b/csmnet/src/C/csm_network_internal_api_c.h
@@ -55,6 +55,7 @@ typedef struct {
     struct sockaddr_un _OtherAddr; // for client sockets this holds the connected addr
 
     char * _DataBuffer;
+    pthread_mutex_t _Lock;
     csm_dgram_buffer_state_t _BufferState;
 
 } csm_net_unix_t;
--- a/csmnet/src/C/csm_network_local.c
+++ b/csmnet/src/C/csm_network_local.c
@@ -467,6 +467,7 @@ int csm_net_unix_Exit( csm_net_endpoint_t *aEP )
 
   if( aEP->_connected )
   {
+    pthread_mutex_lock( &aEP->_ep->_Lock );
     aEP->_connected = 0;
     char *msgData = strdup(CSM_DISCONNECT_MSG);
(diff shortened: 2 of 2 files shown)


lasch:
@fpizzano No idea who to assign for review.

besawn:
@lasch I reviewed the changes in this pull request. It seems like there may be a potential deadlock with the addition of the mutex around the Send/Recv/Shutdown operations, but I am not an expert on how these different operations are triggered so I may be incorrect about this.

The scenario I was considering is: the client is waiting to receive a message when the server terminates the connection to the client. I think that these activities take place in two different threads. So the client thread is blocked waiting to receive when the callback thread receives the message to shut down the connection. In this scenario, I think the client thread will either block until the server is no longer in the disconnected state or perhaps forever. This will cause the callback thread to block while waiting on the mutex held by the client thread. If the lock ever gets freed, it seems like it may no longer be appropriate to shutdown the connection when the callback thread wakes back up.

I'm not sure if my understanding is correct, so the above is all conjecture. Contact me when it is convenient if you would like to discuss in more detail.

lasch:
@besawn Thanks for pointing that out. I don't think there's an actual deadlock. 
The scenario you describe would run into a timeout. And the actual recv timeout (5 seconds) is much shorter than the potentially very long csmi request timeouts.
However, it's still a more serious issue because I noticed that if the disconnect/shutdown code gets the lock before the recv/send, then it cleans up the structures and releases the lock. Then the send/recv goes in and might start operating on the wiped structures.

lasch:
Current approach is a dead end. I'll abandonthe proposed changes and create a new PR.
Closing.

PR #438, opened 2018-12-06
1.4.0 struct finalization

mew2057:
Finalizing the struct versioning for 1.4.0. 

Fixed a minor bug in the error handler.


NickyDaB:
might move the doc file later to better fit in the new doc layout for apis

PR #439, opened 2018-12-06
prevent client lib race condition: socket tear-down and pending send/recv

lasch:
PR solved race condition and deadlock possibility when client lib tears down the link to the CSMD in case of either disconnect msg from daemon or other connection problems.

The approach is to extend the usage of an existing mutex that was supposed to prevent init and term of the lib to interfere. This mutex is now used to make sure termination waits until a recv/send returns and vice versa. It also prevents a recv/send from getting processed while the link data structures get cleaned up.
I believe, a possible deadlock during disconnect existed for a long time, it's just that the probability seemed very low to hit it. The problem was that multiple code paths were possible to enter the disconnect. However, if the main thread performs the disconnect (with the lock aquired) it also terminates the callback thread. If the callback thread attempted to disconnect too, it would wait for the mutex and thus prevents the pthread_join() to complete. With additional use of the mutex it got exposed more and should be fixed with this PR now.

Testing requires to run many iterations of some client cmd (can be error-inject or any other, even if errors are returned like when requesting node attributes of a non-existing node).
While the requests are running, force the local csmd into disconnected mode (e.g. shutting down aggregator for a compute node; disconnect the master of a utility daemon).

The expected result is
* the client output should show errors regarding disconnect of the daemon
* Returning the daemon to regular operation should resume the request iterations.
* the client that got hit by the disconnect should fail but not hang

### DIFF
--- a/csmi/src/common/src/csm_api_common.c
+++ b/csmi/src/common/src/csm_api_common.c
@@ -130,18 +130,31 @@ int csm_init_lib_vers(int64_t version_id)
  */
 int csm_term_lib()
 {
-  pthread_mutex_lock( &disconnect_lock );
   int rc = 0;
+
+  /* cases to handle disconnect/lib termination
+   * - locked+initialized: wait for lock and check initialized again
+   * - unlocked+initialized: get lock and proceed
+   * - locked+uninitialized: return without wait - nothing to do
+   * - unlocked+uninitialized: return without wait - nothing to do */
+
   if (!initialized) {
     csmutil_logging(warning, "not initialized or already terminated\n");
-    pthread_mutex_unlock( &disconnect_lock );
     return ENOTCONN;
   }
-  
+
+  while( pthread_mutex_trylock( &disconnect_lock ) != 0 )
+  {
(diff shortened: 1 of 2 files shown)


lasch:
Still work to do. So far the fixes only reduce the probability. There are still some gaps to close.

lasch:
Some explanation for review:
- the callback thread should not need a lock around its receive call because if the main thread tears down the structs, it will send a thread cancellation and requires the thread to exit before cleanup. Though, I moved the thread cancellation point right after the recv returns so that any cancellation can be handled immediately without attempting to process any received data.
- If the callback thread is doing the cleanup, it will do this via _on_disconnect() which is using the mutex to prevent interference with the main thread.
- The one thing that was found when reviewing with @besawn was the dereference of the struct to call the _on_disconnect() after the mutex got unlocked. This has been resolved by storing the fnptr in a tmp variable.

lasch:
@fpizzano should we merge this one to get more exposure to additional testing?

fpizzano:
Yes, please.

Issue #348, opened 2018-10-18
RPM packaging of python modules required


morrone:
PR #332 introduced a python script called "csm_db_history_archive.py", which is a key component of the csmdb to bds archiving process.  According to the PR:

"The new implementation requires psycopg2 which may be installed using pip: pip install psycopg2"

pip will not be an adequate delivery method.

We require that the psycopg2 module be rpm packaged, and the package that contains the csm_db_history_archive.py script must have a Requires for the psycopg2 package.

Likewise, the "use cases" python scripts depend on the "elasticsearch" python module.  This too needs to be rpm packaged, and the correct Requires need to be added to the rpm that contains the python use case scripts.

It would be very nice to also rpm package the "elasticsearch_dsl" module at the same time.

mew2057:
`psycopg2` is already bundled with red hat. This rpm may be used in lieu of pip.

I need to look into the `elasticsearch` modules.

morrone:
Do we have an rpm dependency on psycopg2 in the mean time?

pdlun92:
Still looks like we are missing the dependency
```
[root@c650mnp05 rpms]# rpm -qR ibm-csm-db-1.4.0-1756.noarch
/bin/sh
/bin/sh
/bin/sh
/bin/sh
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(CompressedFileNames) <= 3.0.4-1
/bin/bash
/bin//bash
/bin/python
rpmlib(PayloadIsXz) <= 5.2-1
```

pdlun92:
Also, looks like there is a `python-elasticsearch` package in epel
```

mattaezell:
FYI @miidon

PR #443, opened 2018-12-07
csmdb_update: 1. Enhancement to the csm db deletion record process - …

williammorrison2:
…created a cleaner script that eliminates wrapper scripts with additional logging features 2. Also included logging tweaks to csm_db_history_archive.py script

## Purpose
To enhance the CSM DB deletion process related to history table records.
Including support of python library.
@morrone included is the python library for postgres dependency #348 in our CMAKE file for RPM build process.

### DIFF
--- a/csmdb/sql/csm_db_history_delete.py
+++ b/csmdb/sql/csm_db_history_delete.py
@@ -2,7 +2,7 @@
 # encoding: utf-8
 #================================================================================
 #   
-#    csm_db_history_archive.py
+#    csm_db_history_delete.py
 # 
 #    © Copyright IBM Corporation 2015-2018. All Rights Reserved
 #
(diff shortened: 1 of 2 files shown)


csmdb/sql/csm_db_history_delete.py
```diff
@@ -0,0 +1,171 @@
...
+# Add handler and formatter.
+handler.setFormatter(LOG_FORMAT)
+logger.addHandler(handler)
+
+# Test the logger create start time and also check the arguments
+a = datetime.datetime.now()
+fullCmdArguments = sys.argv
+argumentList = fullCmdArguments[1:]
+
+if '-n' in argumentList and '-d' in argumentList or \
```
mew2057: This code can be looped into the argparser code.

Issue #507, opened 2019-01-14
csm_transaction.log is too large


morrone:
Our csm_transaction.log on sierra has grown unwieldy at over 36GB in size, and over 5.5million lines.

We need a mechanism to rotate and delete old copies of this log (and others like it) once the data is safe in elasticsearch.

mew2057:
Hi, there should be a rollover mechanism in place today. On the Master Daemon Config file (`/etc/ibm/csm/csm_master.cfg`) modify the `transaction_rotation_size`.

@besawn I mirrored the logging rollover for the standard daemon is there anything you can add on this?

thanh-lam:
Our CSM 1.4.0-1756 has `csm_master.cfg` with "transaction_rotation_size" = 1G. But, I can't tell if it works because the `csm_transaction.log` is not reaching 1G yet. That setting is in the sample `csm_master.cfg` by default, at `/opt/ibm/csm/share/etc/`.

mew2057:
@morrone can you give us your `allocation_metrics_rotation_size` and `transaction_rotation_size` in the `/etc/ibm/csm/csm_master.cfg`.

morrone:
It looks like we have all of these three settings in the "log" area:

```
            "rotationSize"                      : 1000000000,
            "transaction_rotation_size"         : 1000000000,
            "allocation_metrics_rotation_size"  : 1000000000
```

morrone:
So just to be clear, it does not appear that the rotation is working.

thanh-lam:
Agree. It's not working on ours either. I verified it by setting the size to 1000000. And, now `csm_transaction.log` has grown to 166000+. Unless, that setting value means something else.

```
# grep rotation /etc/ibm/csm/csm_master.cfg
            "#rotationSize_comment_1"           :   "Maximum size (in bytes) of the log file, 1000000000 bytes is ~1GB",
            "rotationSize"                      :   1000000000,
            "#transaction_rotation_comment"     :   "Testing on small cluster, default size 1000000000",
            "transaction_rotation_size"         :   1000000,
            "allocation_metrics_rotation_size"  :   1000000

# ls -l csm_transaction.log
-rw-r--r-- 1 root root 1663823 Jan 17 22:28 csm_transaction.log
```

mew2057:
Did some digging, might have a fix ready later toady.

mew2057:
I have a fix ready now, it requires that filebeats change the pattern match for the transaction and metrics logs. PR is pending.

[...]

Issue #502, opened 2019-01-11
CSM failing to report all allocated nodes to BDS


morrone:
It turns out that the list of allocated nodes is being significantly truncated in the CSM's csm_transaction.log.  For instance, allocation 699077 on sierra is allocated 4296 nodes, and using csm_allocation_query we can see the full list of allocated nodes.  However, csm seems to truncate the compute node list to 200 nodes in the csm_transaction.log output.

Since the compute node list is truncated, the kibana csm plugin is overly constraining the search related to allocation 699077.

This make the the plugin not terribly useful.

We need the correct list of compute nodes in the BDS.

We are using csm 1.4.0 on sierra at this time.

mew2057:
Linking to  Issue #498, both bugs have a common issue.

mew2057:
Duplicate of #498

Issue #498, opened 2019-01-10
csm_transaction.log bad JSON in compute_nodes field


morrone:
We still seem to have a lot of bad JSON in the csm_transaction.log.  This is a pretty big issue for us.

Here is a first obvious problem:

We have a transaction of "type":"allocation-step", where the compute_nodes field looks like this:

`"compute_nodes":]`

There is just a closing square brace, but no opening square brace.

This is with csm 1.4.

morrone:
Perhaps not surprisingly, this is also happening in transactions of type "allocation".

mew2057:
What does the compute nodes field look like on the allocation query for an allocation which has this behavior?

morrone:
I just looked at allocation 715416 using csm_allocation_query, which is an allocation with a bad compute_nodes in the csm_transaction.log.  The csm_allocation_query command lists num_nodes as 256, and under compute_nodes there is a list of 256 unique sierra node names.

There are currently 7997 occurances of this in our csm_transaction.log.

This may actually be a longer existing issue, because the first instance of this problem in the csm_transaction log lists the "begin_time" of the allocation as 2018-08-06 18:50:36.227783.
There are 37 instances of this happening thus far today (Jan 10, 2019).

Note that our csm_transaction.log is 36GB in size at this point, and nearly 5.5 million lines.

thanh-lam:
That's helpful to know. Could you also verify that there are other allocations with 256 nodes but didn't have such format problem? I'm thinking maybe there's a limit on how big the compute_nodes field could be. On the clusters I checked, we have only 20-30 nodes max. And, our node names are very short.

morrone:
I did a bit of analysis on a large tail of the csm_transactions.log (my excerpt from the log start on about 2019-12-19, which is significant because it looks like we installed the csm 1.8.0-1756 rpms the day before that).

First, I grep'ed out all of the lines with bad compute_nodes, and then I counted how many of each "num_nodes" size were found.  This is the result:

```
    172 "num_nodes":256
     41 "num_nodes":512
      4 "num_nodes":768
     87 "num_nodes":1024
      6 "num_nodes":1280
      6 "num_nodes":1536
     30 "num_nodes":2048
      4 "num_nodes":2816
      6 "num_nodes":3584
     38 "num_nodes":4096
```

Somewhat interestingly, all of those sizes are a multiple of 256.  But could that be coincidence?

[...]

PR #517, opened 2019-01-18
Fixes to the CSM logs

mew2057:
Fixed JSON generation for transaction log:
Issue #498 
Issue #502

Fixed transaction log rotation and filebeats implementation:
Issue #507

### DIFF
--- a/csmconf/csm_master.cfg
+++ b/csmconf/csm_master.cfg
@@ -12,7 +12,7 @@
             "consoleLog"                        :   false,
             "sysLog"                            :   true,
             "fileLog"                           :   "/var/log/ibm/csm/csm_master.log",
-            "#rotationSize_comment_1"           :   "Maximum size (in bytes) of the log file, 1000000000 bytes is ~1GB",
+            #rotationSize_comment_1"           :   "Maximum size (in bytes) of the log file, 1000000000 bytes is ~1GB",
             "rotationSize"                      :   1000000000,
             "default_sev"                       :   "warning",
             "csmdb"                             :   "info",
--- a/csmconf/csm_master.cfg
+++ b/csmconf/csm_master.cfg
@@ -12,7 +12,7 @@
             "consoleLog"                        :   false,
             "sysLog"                            :   true,
             "fileLog"                           :   "/var/log/ibm/csm/csm_master.log",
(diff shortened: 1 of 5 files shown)


NickyDaB:
What was the original problem? How do the changes here approach to fix this problem?

csmconf/csm_master.cfg
```diff
@@ -12,7 +12,7 @@
             "consoleLog"                        :   false,
             "sysLog"                            :   true,
             "fileLog"                           :   "/var/log/ibm/csm/csm_master.log",
-            "#rotationSize_comment_1"           :   "Maximum size (in bytes) of the log file, 1000000000 bytes is ~1GB",
+            #rotationSize_comment_1"           :   "Maximum size (in bytes) of the log file, 1000000000 bytes is ~1GB",
```
besawn: I am 90% sure that this change breaks the configuration file and prevents the daemon from even starting...

mew2057: I thought this change was removed, fixing the file.

mew2057:
> What was the original problem? How do the changes here approach to fix this problem?
@NickyDaB 

There were two discrete problems:

1. JSON generation was failing for compute nodes where the number was a multiple of 256. This was an issue with the pointer de reference  which was effectively performing a `modulo 256` on the number of nodes. (#502, #498)

2. The transactional (and metrics) log was growing and not rotating. This was resulting in a rather large file on customer systems (#507). This basically indicates that the rollover that had been implemented was non functional.

@besawn I tested these to see if I could get the existing schema to work, but to no good results. I think filebeats may interfere with the clean operation of the boost rollover. This my primary hypothesis for what happened on the customer site, as I couldn't get the traditional rollover to  function.

I adopted the N rollover for two reasons:
1. Preservation of old transactional data in the event of a BDS or filebeat outage. This will allow the transactional tables to be rebuilt wholesale and have no data loss.
2. Circumventing the rollover issue found in [filebeat](https://www.elastic.co/guide/en/beats/filebeat/current/how-filebeat-works.html):


> There is a limitation to Filebeat’s at-least-once delivery guarantee involving log rotation and the deletion of old files. 
> If log files are written to disk and rotated faster than they can be processed by Filebeat, or if files are deleted while the output is unavailable, data might be lost. 
> On Linux, it’s also possible for Filebeat to skip lines as the result of inode reuse. See Frequently asked questions for more details about the inode reuse issue.

PR #519, opened 2019-01-20
Prefer Pre-Incrementing rather than Post-Incrementing

shahzadlone:
Change count++ to ++count to save the additional storage from using the post-increment.

For example consider the following:
```
class Integer {

  public:

  // Sample implementation of pre-incrementing.
  Integer & operator++() {
    myInt += 1;
    return *this;
  }
  // Sample implementation of post-incrementing.
  const Integer operator++(int) {
    Integer temporary = *this;
    ++*this;
    return temporary;
  }

  private:

  int myInt;
};
```

The additional copy (temporary) in post-incrementing step is unnecessary.

### DIFF
--- a/bb/src/BBLV_Metadata.cc
+++ b/bb/src/BBLV_Metadata.cc
@@ -50,7 +50,7 @@ int BBLV_Metadata::update_xbbServerAddData(txp::Msg* pMsg, const uint64_t pJobId
                 for(auto& elements: boost::make_iterator_range(bfs::directory_iterator(job), {}))
                 {
                     LOG(bb,info) << "elements: " << elements;
-                    count++;
+                    ++count;
                 }
                 LOG(bb,info) << "xbbServer: JobId " << pJobId << " is already registered and currently has " << count << " associated job step(s)";
             }


fpizzano:
@shahzadlone Thank you for your interest in this project and creating this PR.

To officially contribute to the project, we do need an CLA signed, if you are covered under a CCLA, please provide that information to the email addresses listed here: https://github.com/IBM/CAST/blob/master/CONTRIBUTING.md, we look forward to any contributions you can make in the future.

fpizzano:
@shahzadlone sorry look like you singed one already. Reopening....

shahzadlone:
Hello, yes i provided that. Thanks.

This should be a pretty easy merge, because it's just a minor change.

shahzadlone:
@fpizzano Do you need anything from me ?

fpizzano:
@shahzadlone I do not, thanks.  @tgooding do you ?

shahzadlone:
Nope there is nothing special about this line. I was going through the source and I found most of the code very beautifully, and well written and then suddenly this one line just really triggered my coding habit / OCD haha. If you want you can point me to those files and i can fix it everywhere. @tgooding

shahzadlone:
This  should be a very trivial merge though.

Issue #520, opened 2019-01-20
Core isolation causes the use of system resources for the application cgroup


peterwywong:
With a new set of 8 nodes, the following two places were checked.

(a) In each of the 8 nodes, /etc/pam.d/sshd had the following line for csmpam where PAM is Portable Authentication Module.
    session    required     libcsmpam.so
    Note that "account    required     libcsmpam.so" is needed according to Fernando. For now, we only had the session line.
(b) In each of the 8 nodes, it has /etc/pam.s/csm/activelist.swp, but not activelist. Fernando said activelist is needed. Again, I'll come back to it later.

Let's focus on ONE level of jsrun first with a microbenchmark of NAS NPB.

This was the sequence of runs:

(1) "ssh launch, step_cgroup"     ===> good
(2) "csm launch, step_cgroup"     ===> good
(3) "csm launch, core isolation"  ===> good

(4) "ssh launch, core isolation"  ===> The resource problem occurred here. Only 8 cores are available for the application.

Based on Fernando's instructions, I did the following:

(A) Added "account    required     libcsmpam.so" BEFORE "session    required     libcsmpam.so" in /etc/pam.d/sshd in each of the 8 nodes.
(B) Added an empty file activelist in /etc/pam.d/csm in each of the 8 nodes.

After this was done, (4) above ran successfully.
I then repeated the above 4 cases with the TWO levels of jsrun (executing jsrun on a compute node). They all worked fine.
Note that before (4), the missing "account    required     libcsmpam.so" line and the missing file activelist did NOT cause failure of the runs.

Now, the questions are:
(i) What caused the resource problem? Why were the resources for the application cgroup and system cgroup flipped after the first core isolation run (3)?
(ii) The two missing things were not needed before the resource problem but needed afterwards. Why?

Probably, a simple sleep program with a request of more than 2 cores per node can reproduce this problem.

SMPI and JSM were built on January 13, 2019.

mew2057:
(i) If I'm reading this correctly this is the expected behavior of the pam module. The pam module was split into an `account` and `session` module with each of them performing different tasks. If both are not present the ssh session will not be placed in the allocation cgroup. I'd need more data captured (`/proc/self/cgroup`) to have a better understanding of exactly what happened in this case.

[...]

PR #521, opened 2019-01-21
Activelist fix

mew2057:
A fix for the issue described in Issue #520 where `activelist.swp` would be found with no `activelist`. New mechanism will create an `activelist` file if able, or rename `activelist.swp` to `activelist`. This new mechanism ensures the list is always present.

Added a fix to the documentation where the `activelist` file was incorrectly named.


thanh-lam:
I verified this fix by removing or renaming the file `activelist` before running a job. For example:
* `mv activelist activelist.swp`
* `mv activelist activelist.other`
* `rm activelist`
In all three cases, the next job ran and a new `activelist` file was created. In the case of `activelist.swp`, this file was removed afterwards.

PR #536, opened 2019-01-26
bb:Robust async request file I/O

dlherms-ibm:
# Pull Request Title

## Purpose
_Describe the problem or feature in addition to a link to the issues._

## Approach
_How does this change address the problem?_

- [ ] Appropriatly label
- [ ] Assign a milestone. 

## How to Test

2. Test it.

### DIFF
--- a/bb/src/bbwrkqmgr.cc
+++ b/bb/src/bbwrkqmgr.cc
@@ -15,6 +15,7 @@
 #include <sys/stat.h>
 #include <sys/syscall.h>
 
+#include <stdio.h>
 #include <string.h>
 
 #include "time.h"
@@ -163,6 +164,7 @@ int WRKQMGR::addWrkQ(const LVKey* pLVKey, const uint64_t pJobId)
 int WRKQMGR::appendAsyncRequest(AsyncRequest& pRequest)
 {
     int rc = 0;
+    size_t l_Size;
 
     LOG(bb,debug) << "appendAsyncRequest(): Attempting an append to the async request file...";
 
@@ -171,7 +173,7 @@ int WRKQMGR::appendAsyncRequest(AsyncRequest& pRequest)
 
   	becomeUser(0, 0);
 
-    int l_Maintenance = NO_MAINTENANCE;
+    MAINTENANCE_OPTION l_Maintenance = NO_MAINTENANCE;
     char l_Cmd[AsyncRequest::MAX_DATA_LENGTH] = {'\0'};
     int rc2 = sscanf(pRequest.getData(), "%s", l_Cmd);
     if (rc2 == 1 && strstr(l_Cmd, "heartbeat"))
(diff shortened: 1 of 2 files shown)


bb/src/bbwrkqmgr.cc
```diff
@@ -181,27 +183,65 @@ int WRKQMGR::appendAsyncRequest(AsyncRequest& pRequest)
...
+                if (l_Size == sizeof(AsyncRequest))
+                {
+                    l_Retry = 0;
+                }
+                else
+                {
+                    // NOTE: If the size written is anything but zero, we are in a world of hurt.
+                    //       Current async request file would now be hosed.
+                    FL_Write(FLAsyncRqst, AppendFail, "Failed fwrite() request, sequence number %ld. Wrote %ld bytes, expected %ld bytes, ferror %ld.", l_SeqNbr, l_Size, sizeof(AsyncRequest), ferror(fd));
+                    if (l_Size)
```
meahoibm: I hate to say it, but should RAS and abort if the weirdness happens.  Very unlikely but if it ever did happen, would be causing problems everywhere.

bb/src/bbwrkqmgr.cc
```diff
@@ -181,27 +183,65 @@ int WRKQMGR::appendAsyncRequest(AsyncRequest& pRequest)
...
+                    {
+                        // Don't retry if a partial result was written
+                        // \todo - What to do...  @@DLH
+                        l_Retry = 0;
+                    }
+                    errorText << "appendAsyncRequest(): Failed fwrite() request, sequence number " << l_SeqNbr << ". Wrote " << l_Size \
```
meahoibm: Consider flightlog of retry?  Suggestion only.

bb/src/bbwrkqmgr.cc
```diff
@@ -592,32 +632,70 @@ int WRKQMGR::findOffsetToNextAsyncRequest(int &pSeqNbr, int64_t &pOffset)
...
+            {
+                threadLocalTrackSyscallPtr->nowTrack(TrackSyscall::fseeksyscall, fd, __LINE__);
+                rc = ::fseek(fd, 0, SEEK_END);
+                threadLocalTrackSyscallPtr->clearTrack();
+                FL_Write(FLAsyncRqst, SeekEnd, "Seeking the end of async request file having seqnbr %ld, rc %ld.", pSeqNbr, rc, 0, 0);
+                if (!rc)
+                {
+                    threadLocalTrackSyscallPtr->nowTrack(TrackSyscall::ftellsyscall, fd, __LINE__);
+                    int64_t l_Offset = -1;
+                    l_Offset = (int64_t)::ftell(fd);
```
meahoibm: Typically would expect a clearTrack() right after the syscall to tightly bracket with the nowTrack.

bb/src/bbwrkqmgr.cc
```diff
@@ -1185,35 +1295,45 @@ FILE* WRKQMGR::openAsyncRequestFile(const char* pOpenOption, int &pSeqNbr, const
...
-                            }
-                            else
-                            {
-                                // Error case...  Just return this file...
+                                    // Iterate to open the new file...
+                                    l_AllDone = false;
+                                }
+                                else
+                                {
+                                    // Error case...  Just return this file...
```
meahoibm: Should maybe log this case.

bb/src/bbwrkqmgr.cc
dlherms-ibm: This is on bbServer...

bb/src/bbwrkqmgr.cc
dlherms-ibm: Retry is too late...  Other bbServer(s) have potentially already read the malformed data and they are now choking also...  In reality, we should restart all bbServers at this point.  Both the general read (with bad read length) was flight logged and the failure flight logged above...

bb/src/bbwrkqmgr.cc
dlherms-ibm: Agreed...  Actually, there is a bug...  We don't untrack in the case where the offset comes back negative...  Will fix this...  Thanks.

bb/src/bbwrkqmgr.cc
dlherms-ibm: Actually, verifyAsyncRequestFile() already logged the failing situation....

dlherms-ibm:
Pull request being closed and being replaced with one that also includes these changes

PR #539, opened 2019-01-28
Bills db updates for 17 0

williammorrison2:
# Bills db updates for 17 0

## Purpose
csmdb update related to the csm_lv_history table
added fields: num_reads, num_writes

- [x] @pdlun92 (regression)
- [x] @NickyDaB code review/test


NickyDaB:
@fpizzano  the new api changes for adding in the 2 new fields are optional and implemented as we discussed. if the field is not supplied then they will be inserted into the database as null.

Issue #371, opened 2018-10-26
Unnecessary extra record on csm_allocation_node_history


fpizzano:
CSM is creating unnecessary extra record on csm_allocation_node_history:

```
csmdb=# select * from csm_allocation_node_history where allocation_id = 103404 and node_name = 'c28n17';
        history_time        | allocation_id | node_name |  state  | shared |  energy   | gpfs_read | gpfs_write |     ib_tx      |     ib_rx      | power_cap | power_shifting_ratio | power_cap_hit
|  gpu_usage  | gpu_energy |    cpu_usage    | memory_usage_max |    archive_history_time
----------------------------+---------------+-----------+---------+--------+-----------+-----------+------------+----------------+----------------+-----------+----------------------+---------------
+-------------+------------+-----------------+------------------+----------------------------
 2018-10-24 11:12:30.524031 |        103404 | c28n17    | failed  | f      |  26412387 |         0 |          0 |  6769001544244 |  6768466059555 |      3050 |                  100 |             0
| 72577443232 |   19980199 | 295262599750514 |      34905784320 | 2018-10-24 11:20:02.157582
 2018-10-24 11:12:30.524031 |        103404 | c28n17    | running | f      |  26412387 |         0 |          0 |  6769001544244 |  6768466059555 |      3050 |                  100 |             0
| 72577443232 |   19980199 | 295262599750514 |      34905784320 | 2018-10-24 11:20:02.157582
 2018-10-24 07:08:56.039039 |        103404 | c28n17    | running | f      | 469223032 | 967905280 | 3584172032 | 63399023857294 | 63503906426527 |      3050 |                  100 |             1
|             |  263971019 |                 |                  | 2018-10-24 07:20:01.624587
(3 rows)

csmdb=#
```

mew2057:
This looks like it's the result of the allocation create/delete workflow and the trigger structure. Working on documenting actions performed on the allocation node.

mew2057:
I think I've tracked the failure point to: `fn_csm_allocation_finish_data_stats`
As it applies to both delete and update events. It looks like the problem was the trigger running twice. My solution is now to disable the trigger and re-enable.

mew2057:
Fix Merged, Closing Issue

PR #542, opened 2019-01-29
Issue #371: Added a mechanism to blink trigger support during stage out.

mew2057:
As per Issue #371 this resolves the bug where two duplicate records are saved to the history table.

To resolve this the `tr_csm_allocation_node_change` is disabled and then re-enabled.


pdlun92:
@mew2057 just a heads up, merge conflict (assuming from #539)

mew2057:
Looks like it was just the changelog.

PR #546, opened 2019-01-31
dev dlherms

dlherms-ibm:
# Pull Request Title

## Purpose
_Describe the problem or feature in addition to a link to the issues._

## Approach
_How does this change address the problem?_

- [ ] Appropriatly label
- [ ] Assign a milestone. 

## How to Test

2. Test it.

### DIFF
--- a/bb/src/bbserver.cc
+++ b/bb/src/bbserver.cc
@@ -1545,7 +1545,7 @@ void msgin_starttransfer(txp::Id id, const string& pConnectionName, txp::Msg* ms
     BBTransferDef* l_TransferPtr = &l_Transfer;
     BBTransferDef* l_OrgTransferPtr = l_TransferPtr;
     bool l_LockHeld = false;
-    bool l_HandleFileLocked = false;
+    HANDLEFILE_LOCK_FEEDBACK l_LockFeedback = HANDLEFILE_WAS_NOT_LOCKED;;
 
     try
     {
@@ -1857,10 +1857,9 @@ void msgin_starttransfer(txp::Id id, const string& pConnectionName, txp::Msg* ms
                                         {
                                             // NOTE: The handle file is locked exclusive here to serialize between this bbServer and another
                                             //       bbServer that is marking the handle/contribid file as 'stopped'
-                                            rc = HandleFile::loadHandleFile(l_HandleFile, l_HandleFileName, l_Job.getJobId(), l_Job.getJobStepId(), l_Handle, LOCK_HANDLEFILE);
+                                            rc = HandleFile::loadHandleFile(l_HandleFile, l_HandleFileName, l_Job.getJobId(), l_Job.getJobStepId(), l_Handle, LOCK_HANDLEFILE, &l_LockFeedback);
                                             if (!rc)
                                             {
-                                                l_HandleFileLocked = true;
                                                 rc = ContribIdFile::loadContribIdFile(l_ContribIdFile, l_HandleFilePath, l_ContribId, l_lvuuid3_Ptr);
                                                 if (rc == 1)
                                                 {
@@ -1941,10 +1940,10 @@ void msgin_starttransfer(txp::Id id, const string& pConnectionName, txp::Msg* ms
                                                 if (rc && l_Continue)
                                                 {
                                                     // Release the lock on the handle file
-                                                    l_HandleFileLocked = false;
-                                                    l_HandleFile->close();
+                                                    l_HandleFile->close(l_LockFeedback);
 
-                                                    if (((wrkqmgr.getDeclareServerDeadCount() - l_Continue) % 15) == 1)
+                                                    uint64_t l_SecondsWaiting = wrkqmgr.getDeclareServerDeadCount() - l_Continue;
+                                                    if ((l_SecondsWaiting % 15) == 1)
                                                     {
                                                         // Display this message every 15 seconds...
                                                         FL_Write6(FLDelay, RestartWaitForStop1, "Attempting to restart a transfer definition for jobid %ld, jobstepid %ld, handle %ld, contribid %ld. Delay of 1 second before retry. %ld seconds remain waiting for the original bbServer to act before an unconditional stop is performed.",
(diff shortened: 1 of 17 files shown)


dlherms-ibm:
This new pull request has all of the changes from pull request #536 (Robust async request file I/O) and the rework for that pull request.  This pull request now also has support for 'old' bbServers to perform stop transfer request processing during failover for transfer definitions they previously owned, but the bbServer has since been restarted.

dlherms-ibm:
@tgooding Let me know how you want to proceed with these changes being merged into master.  I would really like to see robot and IST run with this pull request for awhile before 1.5 is cut.

bb/src/bbserver.cc
```diff
@@ -1954,6 +1953,13 @@ void msgin_starttransfer(txp::Id id, const string& pConnectionName, txp::Msg* ms
...
                                                                      << " seconds remain waiting for the original bbServer to act before an unconditional stop is performed.";
                                                     }
+                                                    // If we will wait for at least another minute, re-append the stop transfer request
+                                                    // every 60 seconds in case the 'old' bbServer just came online...
+                                                    if ((l_Continue > 60) && (l_SecondsWaiting % 60) == 0)
+                                                    {
+                                                        BBLV_Metadata::appendAsyncRequestForStopTransfer(l_TransferPtr->getHostName(), (uint64_t)l_Job.getJobId(),
+                                                                                                         (uint64_t)l_Job.getJobStepId(), l_Handle, l_ContribId, (uint64_t)BBSCOPETRANSFER);
+                                                    }
```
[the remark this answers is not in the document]
dlherms-ibm: Could...  But, the majority of the project is currently written with C-style casts.  Trying to stay consistent with the rest of the code.  The compile time check for these simple integer type casts is not that beneficial.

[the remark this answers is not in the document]
tgooding: Agree about style consistency - its important.  If you were to switch (i'll leave it up to you Dave), it should convert larger blocks of code - and not "update here and there".  I don't have a strong preference toward static_cast, it provides some benefit in narrow spaces.  But don't use dynamic_cast or other RTTI stuff.  Nasty debug scenarios with those.

(It'd be better to standardize the types and avoid the casts completely.  Compilers can generate some suboptimal assembly when types aren't quite identical (wasteful sign-extension is a common one).  But that effort would decidedly not be trivial, better things to do at this stage of development)

bb/src/bbwrkqmgr.cc
```diff
@@ -181,27 +183,65 @@ int WRKQMGR::appendAsyncRequest(AsyncRequest& pRequest)
...
-        pRequest.str(l_Buffer, sizeof(l_Buffer));
-        threadLocalTrackSyscallPtr->nowTrack(TrackSyscall::fwritesyscall, fd, __LINE__, sizeof(AsyncRequest));
-        size_t l_Size = fwrite(l_Buffer, sizeof(char), sizeof(AsyncRequest), fd);
-        threadLocalTrackSyscallPtr->clearTrack();
-        FL_Write(FLAsyncRqst, Append, "Append to async request file having seqnbr %ld for %ld bytes. File pointer %p, %ld bytes written.", l_SeqNbr, sizeof(AsyncRequest), (uint64_t)(void*)fd, l_Size);
-        FL_Write(FLAsyncRqst, CloseAfterAppend, "Close for async request file having seqnbr %ld, file pointer %p.", l_SeqNbr, (uint64_t)(void*)fd, 0, 0);
-        fclose(fd);
-
-        if (l_Size != sizeof(AsyncRequest))
+    int l_Retry = 10;
```
tgooding: The 10 retries should be a #define constant.

bb/src/bbwrkqmgr.cc
```diff
@@ -181,27 +183,65 @@ int WRKQMGR::appendAsyncRequest(AsyncRequest& pRequest)
...
+    while (l_Retry--)
+    {
+        rc = 0;
```
tgooding: Looks like this loop runs as-fast-as-it-can up until 10 retries.  I didn't see any backoff timer/sleeps, does it need one?

bb/src/bbwrkqmgr.cc
```diff
@@ -592,32 +632,70 @@ int WRKQMGR::findOffsetToNextAsyncRequest(int &pSeqNbr, int64_t &pOffset)
...
-    if (fd != NULL)
+    int l_Retry = 10;
+    while (l_Retry--)
     {
-        threadLocalTrackSyscallPtr->nowTrack(TrackSyscall::fseeksyscall, fd, __LINE__);
-        rc = fseek(fd, 0, SEEK_END);
-        threadLocalTrackSyscallPtr->clearTrack();
-        FL_Write(FLAsyncRqst, SeekEnd, "Seeking the end of async request file having seqnbr %ld, rc %ld.", pSeqNbr, rc, 0, 0);
-        if (!rc)
+        rc = 0;
```
tgooding: Same comments as above about backoff timer and define for the 10 retries.

bb/src/bbwrkqmgr.cc
```diff
@@ -592,32 +632,70 @@ int WRKQMGR::findOffsetToNextAsyncRequest(int &pSeqNbr, int64_t &pOffset)
...
+            {
+                threadLocalTrackSyscallPtr->nowTrack(TrackSyscall::fseeksyscall, fd, __LINE__);
+                rc = ::fseek(fd, 0, SEEK_END);
+                threadLocalTrackSyscallPtr->clearTrack();
+                FL_Write(FLAsyncRqst, SeekEnd, "Seeking the end of async request file having seqnbr %ld, rc %ld.", pSeqNbr, rc, 0, 0);
+                if (!rc)
+                {
+                    threadLocalTrackSyscallPtr->nowTrack(TrackSyscall::ftellsyscall, fd, __LINE__);
+                    int64_t l_Offset = -1;
+                    l_Offset = (int64_t)::ftell(fd);
```
tgooding: Unrelated to this code change.  But ftell returns long, not int64_t.  There is also a ftello() that returns off_t.

bb/src/bbwrkqmgr.cc
```diff
@@ -694,39 +773,67 @@ int WRKQMGR::getAsyncRequest(WorkID& pWorkItem, AsyncRequest& pRequest)
         l_SeqNbr -= 1;
     }
 
-    FILE* fd = openAsyncRequestFile("rb", l_SeqNbr);
-    if (fd != NULL)
+    int l_Retry = 10;
+    while (l_Retry--)
```
tgooding: Same as above...

tgooding:
Few minor things, but looked good to me.

bb/src/bbwrkqmgr.cc
dlherms-ibm: Agreed....  Actually that '10 retry' is used in a few places.  Next set of changes, I'll incorporate that change...

bb/src/bbwrkqmgr.cc
dlherms-ibm: Even with the code that does run, it might be worth while to still add a delay.  We are already in strange territory where an fwrite is failing....  Not sure if it will work after ten tries...  I'll add a delay in future work...

bb/src/bbwrkqmgr.cc
dlherms-ibm: Will do...

bb/src/bbwrkqmgr.cc
dlherms-ibm: I can make that more explicit with the types...  Thanks.

bb/src/bbwrkqmgr.cc
dlherms-ibm: Will do...

PR #548, opened 2019-01-31
Fix dynamic_date_formats bug in elasticsearch templates

### LINKED ISSUE #547: elasticsearch templates with dynamic_date_formats are broken
All of the templates that set dynamic_date_formats have a formatting error and are rejected by elasticsearch.  I'll submit the fix.

morrone:
In dynamic_date_formats "strict_date_optional_time" needs to be a
separate comma-separated option, not or'ed with the date time patterns.

Fixes #547


morrone:
It is also possible that double bars would have worked.  I didn't try that.

Issue #222, opened 2018-08-21
CSM lacks collective commands


fpizzano:
**Summary:**
Another issue with CSM that we have noticed. I assume that this has already been logged, but perhaps worthwhile for LLNL to log as well. 

CSM lacks collective commands, such as the ability to specify sierra[100-360] -r y, to make ready a large number of nodes. Without such collective commands, we will have to wrap scripts CSM API binaries, write SQL scripts, etc This may have performance implications as well, if we have to make ready many nodes. 

It''s nnclear if we are missing something, or if this functionality indeed doesn''t exist. 

**Is your feature request related to a problem? Please describe.**
Problem Report number: 37974

fpizzano:
Adding @adambertsch  @watson6282

mew2057:
A high level assessment of this feature makes me think this could be a achieved through a multi phase process on the master daemon:

1. Check if xCAT is running on the node. 
2. If xCAT is present query the group/range from xcat.
3. Continue to use this generated range in the rest of the API.

I'm not sure if this is an acceptable mechanism as it will likely add a database call to every query with an xCAT range or group. It might be best to limit it just to range and look at xCAT's range specifiers re-implementing them in C++.

NickyDaB:
@gurevichmark sent me this. posting it here for later reference. 

https://xcat-docs.readthedocs.io/en/stable/advanced/restapi/restapi_resource/restapi_reference.html#uri-nodes-noderange-nodels-lists-the-nodes-noderange-cannot-start-with

NickyDaB:
I read the above link in more detail. seems promising.

NickyDaB:
More info. For python:
https://xcat-docs.readthedocs.io/en/stable/advanced/restapi/restapi_usage/restapi_usage.html#an-example-of-how-to-use-xcat-rest-api-from-python

NickyDaB:
Example of api in action in python.

https://github.com/xcat2/xcat-core/blob/master/xCAT-server/xCAT-wsapi/xcatws-test.py

NickyDaB:
@pdlun92 

@gurevichmark  says we have an old version of xcat

```
[root@c650mnp02 nbuonar]# lsxcatd -a
Version 2.13.7
```

still looking into it, but we may need to upgrade to get the noderange support.

NickyDaB:
```
[...]

PR #555, opened 2019-02-06
an example of how to use xCAT restful python APIs and CSM python APIs to use an xCAT noderange

NickyDaB:
# an example of how to use xCAT restful python APIs and CSM python APIs to use an xCAT noderange

## Purpose
We want to be able to use an xCAT noderange or nodegroup with CSM APIs.

## Approach
This approach uses the python wrapper for CSM APIs. By writing a python script we can first call xCAT python restful APIs to get the nodes in a node range, then save that and pass it into the input of a CSM API, such as `node attributes query`. 

## Contents
This pull request includes :
- an example of how to use xCAT restful python APIs and CSM python APIs to use an xCAT noderange
- An update to the FAQ section of CSM python APIs read the docs to include a note about utf8 encoding. 

## How to Test
1. Nothing to new to test. This pull request add in an example python script of how to use an xCAT node range. not meant to be used in production.
2. Example works and has been tested.

## Origin:
original issue: https://github.com/IBM/CAST/issues/222

### DIFF
--- a/csmi/python_samples/node_attributes_query_with_xcat.py
+++ b/csmi/python_samples/node_attributes_query_with_xcat.py
@@ -2,9 +2,9 @@
 # encoding: utf-8
 #================================================================================
 #
-#    node_attributes_query.py
+#    node_attributes_query_with_xcat.py
 #
-#  © Copyright IBM Corporation 2015-2018. All Rights Reserved
+#  © Copyright IBM Corporation 2015-2019. All Rights Reserved
 #
 #    This program is licensed under the terms of the Eclipse Public License
 #    v1.0 as published by the Eclipse Foundation and available at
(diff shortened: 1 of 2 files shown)


csmi/python_samples/node_attributes_query_with_xcat.py
```diff
@@ -0,0 +1,124 @@
...
+# Here is your xcat search. 
+# replace as needed for your search. 
+# examples:
+# nodes - c650f03p[07,09,11]
+# node range - c650f03p[07-10]
+# node group - all
+node_string = "c650f03p[07,09,11]"
+
+# use the xCAT restAPI url to construct a message that will return the information you want.
+get_node_range = REST_ENDPOINT + "/nodes/" + node_string + "/nodels/"
```
mew2057: More pythonic might be:

``` python
 get_node_range ='{0}/nodes/{1}/nodels/'.format(REST_ENDPOINT, node_string)
```

This is not necessary, but might be useful in the future.

NickyDaB: oh interesting. didn't know python worked liked that

NickyDaB: since this is an example to show the systems interacting, i'll leave alone. but I'll use this in future scripts that I make for production.

PR #566, opened 2019-02-11
Improving default configuration of PAM module

mew2057:
Learned of some new configuration implications of the PAM module. RPM install script now configures  the module about the `password-auth` for the account.


mew2057:
1. I think I'm going to remove the readme, because it's now superseded by the ReadTheDocs documentation and I don't want to maintain two readmes for the foreseeable future.

2. Sure, I might even be able to knock a quick unit/regression up this AM since I'm in planning/remediation mode.

PR #604, opened 2019-03-08
copyright fix 2019

williammorrison2:
copyright header fix

### DIFF
--- a/csmdb/sql/csm_db_utility_setup_script.sh
+++ b/csmdb/sql/csm_db_utility_setup_script.sh
@@ -3,7 +3,7 @@
 #
 #    csm_db_utility_setup_script.sh
 #
-#  © Copyright IBM Corporation 2018. All Rights Reserved
+#  © Copyright IBM Corporation 2019. All Rights Reserved
 #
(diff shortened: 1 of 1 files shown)


lasch:
I know this is closed already. However, in case you want the copyright year automatically checked the moment you type: `git commit ...` you might consider adding this snippet to `.git/hooks/pre-commit` in your local sandbox (create the file if it doesn't exist or copy the `pre-commit.sample` first).

It will not check for the copyright header itself, instead it checks if there is already a copyright header and if so, it checks if the current year is part of it.

```
year=$(date '+%Y')
filelist=$(git status --porcelain | grep "^[MADRCU]" | awk '{print $2}')
for file in ${filelist}
do
    head -n 10 ${file} | grep "Copyright" &> /dev/null || continue
    head -n 10 ${file} | grep "Copyright.*${year}" &> /dev/null && crfine=0 || crfine=1
    if [ $crfine -eq 1 ]
    then
        echo
        echo "!!!  ERROR: Copyright year ${year} is missing in header of ${file} !!!"
        echo
        exit 1
    fi
done
```
If you find problems or have suggestions for improvement, let me know.

PR #609, opened 2019-03-13
Removing python-elasticsearch dependency

mew2057:
The python-elasticsearch dependency was erroneously added. The version present in the rhels 7.5 epel is far too old for the BDS offering. A requirements.txt file will now be provided with the CSM BDS usecases to handle dependencies.


fpizzano:
Adding Py and Adam @watson6282 @adambertsch

fpizzano:
Adding @miidon and @mattaezell

PR #614, opened 2019-03-14
csmdb_python_dependency

williammorrison2:
# Purpose
Python dependency (add appropriate dependencies to the CSM rpms so yum can automatically install all of the right packages.)

- [x]  @pdlun92 to review (regression)
- [x] @besawn review
- [ ] @mew2057 review
- [ ] @fpizzano  review

### DIFF
--- a/csmd/setupRPM.cmake
+++ b/csmd/setupRPM.cmake
@@ -30,6 +30,7 @@ set( CPACK_RPM_csm-core_POST_INSTALL_SCRIPT_FILE
 # ibm-csm-db rpm settings
 set(CPACK_RPM_csm-db_PACKAGE_ARCHITECTURE "noarch")
 set(CPACK_RPM_csm-db_PACKAGE_REQUIRES "pv")
+set(CPACK_RPM_csm-db_PACKAGE_REQUIRES "python-psycopg2")
 
 # ibm-csm-hcdiag rpm settings
 set(CPACK_RPM_csm-hcdiag_PACKAGE_ARCHITECTURE "noarch")
--- a/csmd/setupRPM.cmake
+++ b/csmd/setupRPM.cmake
@@ -29,8 +29,7 @@ set( CPACK_RPM_csm-core_POST_INSTALL_SCRIPT_FILE
 
 # ibm-csm-db rpm settings
 set(CPACK_RPM_csm-db_PACKAGE_ARCHITECTURE "noarch")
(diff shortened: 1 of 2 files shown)


csmd/setupRPM.cmake
```diff
@@ -30,6 +30,7 @@ set( CPACK_RPM_csm-core_POST_INSTALL_SCRIPT_FILE
 # ibm-csm-db rpm settings
 set(CPACK_RPM_csm-db_PACKAGE_ARCHITECTURE "noarch")
 set(CPACK_RPM_csm-db_PACKAGE_REQUIRES "pv")
+set(CPACK_RPM_csm-db_PACKAGE_REQUIRES "python-psycopg2")
```
tgooding: Did you intend to replace the "pv" package dependency with "python-psycopg2"?  Or should it be "pv,python-psycopg2"?

williammorrison2: @tgooding we would be including both pv and python-psycopg2.

tgooding: ok, looks like the second set() is dropping the pv dependency then.

williammorrison2: @tgooding can you combine the two? Or is there an alternative solution that you would suggest?

pdlun92:
Passes regression and fixes the issue:
rpm dependencies for current daily master build (without fix):
```
# rpm -qpR  /ugadmin/CORAL/1.1.0/CORAL_Product/production/latest_images_ppc64le_rh75/ibm-csm-db-1.6.0-2219.noarch.rpm
pv
/bin/sh
/bin/sh
/bin/sh
/bin/sh
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(CompressedFileNames) <= 3.0.4-1
/bin/bash
/bin//bash
/bin/python
rpmlib(PayloadIsXz) <= 5.2-1
```
rpm dependencies for build from this pull-request
```
# rpm -qR ibm-csm-db-1.6.0-2222.noarch
pv
python-psycopg2
/bin/sh
/bin/sh
/bin/sh
/bin/sh
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(CompressedFileNames) <= 3.0.4-1
/bin/bash
/bin//bash
/bin/python
rpmlib(PayloadIsXz) <= 5.2-1
```

PR #619, opened 2019-03-20
Fix:  Report to bhist status of stageout

meahoibm:
Problem:
When a failure during stage out (injected error in these cases), the stage out fails (as expected) but that error is not percolated back to the user.
Provide status of stageout to bhist.


meahoibm:
See 
```failure during stage-out does not report error to user #1019```
for test result examples.

tgooding:
Looks good

PR #629, opened 2019-03-29
Added interfacename flag for case where ib0 is not the interface name…

foxdllnl:
# Purpose
Allow for alternate name of interface used with NVMEoF

# Approach
New flag interfacename with default being original hard coded value (ib0)


tgooding:
Looks good, thanks Dave!

PR #636, opened 2019-04-05
Fixing debug level for PAM scripts

mew2057:
The debug level `off` is incorrectly named in the context of the logging utility. Setting the logging level to `disabled`.

New library retested and no output was found.


fpizzano:
Adding @adambertsch

mew2057:
@pdlun92 can we run a pull request and an ssh to a pam configured node with an active allocation? I can't repeat the original bug in my environment and I think it might have something to do with the  weirdness of my config.

PR #646, opened 2019-04-11
Csm ufm compatibility -> master

NickyDaB:
# Purpose
- new ufm restAPIs had compatibility issues with CSM. this pull request addresses those issues and adds backwards compatibility and new testing capabilities. 

- also, the new ufm includes values for previous fields that were blank. CSM now takes these values and inserts them into CSM database. 

# Approach
- We added layers of abstraction to the process to use more things through fields and variables, rather than assuming formats and hard coding things. Because of this, some new fields were added to `master.cfg`
- updated read the docs to include notes on changes and upgrades included in pull request
- added some include files for csmtest, to help with setting up some automated testing

# Origin
found via manual tests

# How to Test
- [x] new features have been documented in the `-h` and read the docs, but @pdlun92 should meet with @NickyDaB offline and discuss plans to integrate automated testing. 
- [x] add a check for loading 5_9 data and see if program crashes
- [x] add a check for loading 6_1 data and see if program crashes
- [x] add a check for 6_1 data to see if the `hw_version` field is correctly filled inside a record of `csm_switch_inventory` table, in 5_9 the field should be blank in csm database because ufm did not provide it.

PR #647, opened 2019-04-11
Nick cast 1.5.x -> IBM 1.5.x

NickyDaB:
i think this is all the appropriate cherry picks. should be same as other pull request. 

:

cherry picks to 1.5.x only


mew2057:
I'm really confused as to why there are two PR with mostly the same commits. Additionally, despite the two commits being to separate branches there's missing commits in the one going to master (#646).

Are the other commits needed in master? This seems to break with our previous strategy of branch  resolution.

PR #649, opened 2019-04-18
Check in preuninstall whether upgrade or delete of export_layout

meahoibm:
Handles uninstall of export layout rpm and upgrade.
Previously, upgrade removed export_layout module.
This will require two successive RPM installs--first with previous export_layout RPM removed.  The second RPM install will work for rpm -U.


tgooding:
Looks good, thanks!

PR #651, opened 2019-04-24
Stderrjson

meahoibm:
bbcmd adds csm stderr i/o to json when csm API is called.


bb/src/bbGrabStderr.cc
```diff
@@ -0,0 +1,64 @@
...
+ |
+ |    U.S. Government Users Restricted Rights:  Use, duplication or disclosure
+ |    restricted by GSA ADP Schedule Contract with IBM Corp.
+ *******************************************************************************/
+#include <stdio.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include "bbGrabStderr.h"
+
+    GrabStderr::GrabStderr(int pBuffSize){
```
tgooding: buffSize should be size_t.

bb/src/bbGrabStderr.cc
```diff
@@ -0,0 +1,64 @@
...
+        buffEnd = 0;
+        pipe2(stderrPipeFD, O_CLOEXEC);
+        dupSTDERR_FILENO = dup(STDERR_FILENO);
+        dupRC = dup2(stderrPipeFD[1],STDERR_FILENO);
+        setNonBlockMode(stderrPipeFD[0]);
+        setNonBlockMode(stderrPipeFD[1]);
+        setNonBlockMode(STDERR_FILENO);
+    }
+
+    char * GrabStderr::getStdErrBuffer(){
```
tgooding: Seems to me that the pointer to the buffer should be passed into getStdErrBuffer() rather than allocated in the constructor and pointer returned by getStdErrBuffer().  The later is a bit unsafe as the destructor could reclaim storage.

Issue #658, opened 2019-04-26
Premature CSM timeout during csm_bb_cmd


tgooding:
The `alltests.lsf.Tests.Lsf.Basic.Load job queue with BB` testcase has been failing.  Debugged it a bit and it appears that csm_bb_cmd is returning with a "Timeout Occured in State: 2" after only ~2.5 milliseconds.  

This is a testcase that loads the LSF queue with BB jobs and expects them to complete normally.  So there will be jobs overlapping (e.g., parts of the job in stage-out phase while another job is either allocating or running).  It may be similar to behavior seen on butte.

Full csm_master log on c650mnp03-ug:/var/log/ibm/csm/csm_master.log.  But here is (what I think is) a relevant snippet:
```
2019-04-26 03:05:20.128307     csmapi::info     | [1053110605]; csm_allocation_create start
2019-04-26 03:05:20.128408     csmapi::info     | [1053110605]; Primary Job Id: 128438; Secondary Job Id: 0; Message: Requesting allocation in database; 
2019-04-26 03:05:20.131348     csmapi::info     | [1053110605]; Allocation ID: 3; Primary Job Id: 128438; Secondary Job Id: 0; Message: Reserving nodes in database; 
2019-04-26 03:05:20.131783     csmapi::info     | [1364333597]; csm_allocation_query_active_all start
2019-04-26 03:05:20.132611     csmapi::info     | [1364333597]; csm_allocation_query_active_all end
2019-04-26 03:05:20.134888     csmapi::info     | csm_allocation_create[1053110605]; Allocation ID: 3; Primary Job Id: 128438; Secondary Job Id: 0; Message: Create Completed; 
2019-04-26 03:05:20.135610     csmapi::info     | [1053110605]; csm_allocation_create end
2019-04-26 03:05:20.136748     csmapi::info     | [1694795309]; csm_allocation_create start
2019-04-26 03:05:20.136953     csmapi::info     | [1694795309]; Primary Job Id: 128439; Secondary Job Id: 0; Message: Requesting allocation in database; 
2019-04-26 03:05:20.137130     csmapi::info     | [653532070]; csm_allocation_create start
2019-04-26 03:05:20.137255     csmapi::info     | [653532070]; Primary Job Id: 128440; Secondary Job Id: 0; Message: Requesting allocation in database; 
[...]

PR #665, opened 2019-05-02
csm: fix for #658

lasch:
set variable according to interrupted sleep status to prevent premature event trigger.
issue #658 was caused by missing variable assignment that was needed to restore a timer event after an interrupted sleep. This problem does not exist in 1.5 or earlier releases. It was introduced by commit 2905ab9 
This PR should fix the problem.

I also made some minor changes to reduce the number of system clock reads (in case logging was enabled).

Regarding testing to catch situations like that in the regression buckets:
 - we'd need to find a way to interrupt currently active sleeps and confirm correct behavior
 - one way could be to use the error_inject test tool: the test case for timeout testing uses a timer event and is therefore a candidate for testing.
   1) request a timeout test with maybe 10s
   2) send a signal to the master daemon that causes sleep to be interrupted (need to look/experiment which signals 
   3) make sure the timeout test takes at least 10s

### DIFF
--- a/csmd/src/daemon/src/csm_timer_manager.cc
+++ b/csmd/src/daemon/src/csm_timer_manager.cc
@@ -55,11 +55,12 @@ void TimerManagerMain( csm::daemon::EventManagerTimer *aMgr )
        */
       csm::daemon::TimerContent content = tev->GetContent();
 
+      int64_t remaining = content.RemainingMicros();
       // uSleep does an interruptable sleep and returns true if we got interrupted
-      LOG( csmd, trace ) << "Setting timer to trigger in " << content.RemainingMicros() << "µs.";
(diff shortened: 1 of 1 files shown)


lasch:
Tried to create a test with sending signals to csmd and that is not working in a useful way.
The other (probably better) option is to cause an interleaved set of timers. Whenever a new timer event is added and it's expiration is before the currently earliest timer, this timer is interrupted. And in the unfixed version it would cause the new timer to trigger the old one prematurely.

Here's how I was able to test this:
 - take a master daemon and configure `csm.recurring_tasks.soft_fail_recovery.interval : 00:00:15`
 - start the daemon
 - run `time error_inject -t 20`

Without the fix, almost all error_inject calls will return earlier than 20s.
With the fix, all calls should return after 20s.

Issue #679, opened 2019-05-10
Too many json files in /var/log/ibm/csm/archive


morrone:
There are far too many *.json files in /var/log/ibm/csm/archive, and I'm now hitting a limit in filebeat.  I can see if there is a way to increase the limit, but it really doesn't make sense to have filebeat monitoring old files that are not intended to every have new data.

We need some strategy to improve this situation.

thanh-lam:
I just looked on my cluster. There are lots of empty (0 file size) files. Every time the archive command is run, many db tables don't have any activities in their history. Again, maybe that's just on my cluster. In this case, the first thing I'd do is to remove all the empty files to reduce the number of files. But, it's not going to save any space.

morrone:
Yes, there are many empty files.  CSM should probably stop doing that.

I'm less concerned about space in this ticket than I am about the shear number of files.  The large number (over 5000) is causing issues already.

williammorrison2:
@morrone we are working on a solution to rectify this current situation.

mew2057:
I have implemented a first pass fix for this issue. I've implemented a script which collates existing json into weekly slices. We might want to make the baseline archive script slice the data into weekly files as well, removing the need for the script in the future. Additionally, I've modified the filebeat to close files after they've been moved or set to inactive.

besawn:
I think this was resolved via #688, please reopen if you are still experiencing this issue.

PR #688, opened 2019-05-17
Reducing number of json files #679

mew2057:
# Purpose
A Fix for the too many JSON file problem.

# Approach
Implemented a script to consolidate the JSON files.

Added new configuration to filebeat to handle file closures. Implemented `csm_db_rollup.py` to roll old files up into weekly files.

This doesn't fix the files produced or empty files yet. @williammorrison2 Should we switch to a file containing weekly data for the archive?


mew2057:
@williammorrison2 I implemented the real time fix we talked about and added some UTF-8 string sanitization to the pull process.

Issue #681, opened 2019-05-14
Error and warning during CSM BDS update


thanh-lam:
**Describe the bug**
With CSM 1.6, the bds-* rpms can be updated to a new level. Even though the update was completed, there were some error and warning messages:
"/var/tmp/rpm-tmp.pXOWzC: line 8: syntax error: unexpected end of file"

**To Reproduce**
1. Check to be sure the current bds level: `rpm -qa | grep csm`.
2. Stop the kibana and elasticsearch services.
3. Run `yum update ibm-csm-bds-<...>` for bds, bds-kibana, bds-logstash.
4. Watch for error or warning between the displayed messages.

**Expected behavior**
Update should have no such warning or error.

**Screenshots**
```
...
Cast Visulizer added to kibana plugins.
  Updating   : ibm-csm-bds-logstash-1.6.0-2444.noarch                                       3/6 
Adding Event Correlator to logstash plugins.
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 51876790272 bytes for committing reserved memory.
# An error report file with more information is saved as:
# //hs_err_pid4233.log
OpenJDK 64-Bit Server VM warning: TieredCompilation is disabled in this release.
OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00007ff207e70000, 51876790272, 0) failed; error='Cannot allocate memory' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 51876790272 bytes for committing reserved memory.
# An error report file with more information is saved as:
# //hs_err_pid4337.log
Event Correlator has been added to logstash plugins.
  Cleanup    : ibm-csm-bds-1.6.0-2337.noarch                                                4/6 
  Cleanup    : ibm-csm-bds-kibana-1.6.0-2337.noarch                                         5/6 
Removing Cast Visulizer from kibana plugins.
/var/tmp/rpm-tmp.pXOWzC: line 8: syntax error: unexpected end of file
warning: %postun(ibm-csm-bds-kibana-1.6.0-2337.noarch) scriptlet failed, exit status 2
Non-fatal POSTUN scriptlet failure in rpm package ibm-csm-bds-kibana-1.6.0-2337.noarch
  Cleanup    : ibm-csm-bds-logstash-1.6.0-2337.noarch                                       6/6 
[...]

PR #696, opened 2019-05-21
Repaired ibm-csm-bds-kibana pre.uninstall scriptlet

mew2057:
Scriptlet had malformed bash script. Fixed if statement.


mew2057:
Screwed up reference in commit log should be #681

PR #722, opened 2019-06-13
Do findSerials early and remove pthread_once invocations

meahoibm:
Removed pthread_once of findSerials and placed priming call in mainThread before worker and transfer thread creates.

Unit tested on gssiovm1-hs for good path with existing nvme-over-fabrics devices.  

Unit tested on gssiovm1-hs for bad path of ```chmod a-x /usr/sbin/nvme```


tgooding:
Looked good.  Merging.

PR #759, opened 2019-07-18
CSM Allocation Multiple Allocation fix

mew2057:
A fix for Allocations with multiple LSF job ids that match.


mew2057:
@williammorrison2 @besawn  Ready for test.

csmi/include/struct_generator/generate_struct.pl
```diff
@@ -316,7 +316,7 @@
 # Else build just the alias.
 if ( $alias_name  eq "" )
 {
-    $struct .=  "typedef struct {${NFIELDS}${struct_def}} ${struct_name};";
+    $struct .= "typedef struct ${struct_name} ${struct_name};\nstruct ${struct_name} {${NFIELDS}${struct_def}};";
```
besawn: This change ends up generating a minor change to the boiler plate of every CSM API struct. Can provide some more context on why this is included? Is this something that was flagged by coverity, for example, or is it needed to support the changes for the CSM Allocation Multiple Allocation fix?

mew2057: You need the typedef for the circular reference.

Issue #760, opened 2019-07-18
CSM FVT: Test Case for csm_allocation_query more than 1 allocation with job_id=1


williammorrison2:
**Summary:**
Related to PR #759 
The test case(s) may need some redesign based on the new "multiple allocations per job id" behavior.

fpizzano:
Current behavior from the current master build:

```
[root@c650f02p07 ~]# head -1 /var/log/ibm/csm/csm_utility.log
[UTL]2019-07-24 10:25:54.498502       csmd::info     | CSMD: UTILITY; Version: 1.7.0; Build: 718f5d372e; Date: Sun Jul 21 23:54:58 2019 -0500; Config file: /etc/ibm/csm/csm_utility.cfg
[root@c650f02p07 ~]#
```

```
[root@c650f02p07 ~]# /opt/ibm/csm/bin/csm_allocation_query -j 101
[csmapi][warning]	/home/ppsbld/workspace/PUBLIC_CAST_V1.0.0_ppc64LE_RH7.5_ProdBuild/csmi/src/common/src/csmi_common_utils.c-147: the Error Flag Set
[csmapi][error]	csmi_sendrecv_cmd failed: 25 - csm_allocation_query[1963466001]; Too many allocations matching the query were found, something has gone wrong.
/opt/ibm/csm/bin/csm_allocation_query FAILED: errcode: 25 errmsg: csm_allocation_query[1963466001]; Too many allocations matching the query were found, something has gone wrong.
[root@c650f02p07 ~]#
```

```
csmdb=# select * from csm_allocation_history where primary_job_id = 101;
        history_time        | allocation_id | primary_job_id | secondary_job_id | ssd_file_system_name | launch_node_name | isolated_cores | user_flags | system_flags | ssd_min | ssd_max | num_nodes
 | num_processors | num_gpus | projected_memory |  state   |     type     | job_type | user_name | user_id | user_group_id | user_group_name | user_script |         begin_time         |          end
_time          | exit_status | account | comment | job_name |   job_submit_time   | queue | requeue | time_limit | wc_key | archive_history_time | smt_mode | core_blink
----------------------------+---------------+----------------+------------------+----------------------+------------------+----------------+------------+--------------+---------+---------+----------
-+----------------+----------+------------------+----------+--------------+----------+-----------+---------+---------------+-----------------+-------------+----------------------------+-------------
---------------+-------------+---------+---------+----------+---------------------+-------+---------+------------+--------+----------------------+----------+------------
[...]

PR #767, opened 2019-07-24
added test case regarding #759 and #760

williammorrison2:
# Purpose
added test case regarding #759 and #760

Ran regression and the specific test case passes along with all other expected behavior.

- [x] @pdlun92 to review (code/test case)
- [x] @besawn to review

### DIFF
--- a/csmtest/buckets/error_injection/allocation.sh
+++ b/csmtest/buckets/error_injection/allocation.sh
@@ -157,11 +157,11 @@ check_return_flag_nz $? 4 "Test Case 14: csm_allocation_query job id does not ex
 
 # Test Case 15: csm_allocation_query more than 1 allocation with job_id=123
 ${CSM_PATH}/csm_allocation_query -j 123 > ${TEMP_LOG} 2>&1
-check_return_flag_nz $? 0 "Test Case 15: csm_allocation_query more than 1 allocation with job_id=123"
+check_return_flag $? "Test Case 15: csm_allocation_query more than 1 allocation with job_id=123"
 
 # Test Case 15: csm_allocation_query more than 1 allocation with job_id=123 - verify num_allocations
 check_all_output "num_allocations: 4"
-check_return_flag_nz $? 0 "Test Case 15: csm_allocation_query more than 1 allocation with job_id=123 - verify num_allocations"
+check_return_flag $? "Test Case 15: csm_allocation_query more than 1 allocation with job_id=123 - verify num_allocations"
 
 # Test Case 16: csm_allocation_query invalid -a option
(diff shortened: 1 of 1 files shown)


csmtest/buckets/error_injection/allocation.sh
```diff
@@ -155,9 +155,13 @@ check_return_flag_nz $? 4 "Test Case 13: csm_allocation_query allocation does no
 ${CSM_PATH}/csm_allocation_query -j 123456789 > ${TEMP_LOG} 2>&1
 check_return_flag_nz $? 4 "Test Case 14: csm_allocation_query job id does not exist"
 
-# Test Case 15: csm_allocation_query more than 1 allocation with job_id=1
-${CSM_PATH}/csm_allocation_query -j 1 > ${TEMP_LOG} 2>&1
-check_return_flag_nz $? 0 "Test Case 15: csm_allocation_query more than 1 allocation with job_id=1"
+# Test Case 15: csm_allocation_query more than 1 allocation with job_id=123
+${CSM_PATH}/csm_allocation_query -j 123 > ${TEMP_LOG} 2>&1
+check_return_flag_nz $? 0 "Test Case 15: csm_allocation_query more than 1 allocation with job_id=123"
```
besawn: An additional question for @pdlun92:
I think these two checks in Test Case 15 for `check_return_flag_nz $? 0` will function correctly, but it seems to deviate from the general FVT test case style. In general, it seems like `check_return_flag $?` is used when the expected return value is 0 and  `check_return_flag_nz $? EXPECTED_VALUE` is used when the expected return value is anything other than 0.

Would you recommend that these two checks be changed from:
`check_return_flag_nz $? 0`
to
`check_return_flag $?`

Should we simplify things by replacing both of these functions with some new function called check_return_flag_value (or some other name) and also explicitly use it with the expected return value instead of the two existing functions?

We can discuss in more detail at the next team meeting, just curious about your thoughts here.

pdlun92: Good catch.  And I agree, moving forward we may want to converge in to one function.  Definitely a good topic for next team meating

pdlun92:
> For my own education: what is the significance of changing the jobid for test cases 2,3,4, and 15 from jobid=1 to jobid=123?

@besawn two driving reasons behind that.  We wanted to isolate some of these allocations from the rest of the `complete_fvt`, which uses job ID 1.  We isolated these allocations so that this query test case works both in a vacuum and in complete regression.  Second, with this script up to test case 15 there are 4 allocations with job=123, 3 of which are failed allocations and 1 which is active.  We designed it like this to verify that the API is able to grab multiple allocations from both active and history tables

PR #769, opened 2019-07-25
Refactor tracksyscall to be able to monitor runcommand

meahoibm:
Refactor of tracksyscall in bb into tracksyscall in utilities so that runcommand in utilities can be monitored for taking too long.  The monitor is tracksyscallMonitor.cc in bb/src.

Need review.

Have unit tested and removed unit test defects.


tgooding:
Lets hold this PR until 1.6.1 changes are cherry-picked.

tgooding:
@dlherms-ibm : can you take a look?  Thanks!

utilities/src/tracksyscall.cc
```diff
@@ -0,0 +1,56 @@
...
+    _syscall = nosyscall;
+    _timeStamp=0;
+    _lineNumber=0;
+    _tid=-1;
+    _fd=-1;
+    _fileName.clear();
+}
+
+TrackSyscall::TrackSyscall()
+{
```
dlherms-ibm: _filename not explicitly inited and _timestamp inited twice...

bb/src/tracksyscallMonitor.cc
```diff
@@ -84,11 +67,9 @@ void checkForStuckSyscall()
     uint64_t nowTimeStamp = timeStamp();
     uint64_t tick_diff = 0;
 
-    pthread_mutex_lock(&tidTrackerMutex);
-    auto l_pthread_syscalltracker = pthread_syscalltracker; //make a local copy
-    pthread_mutex_unlock(&tidTrackerMutex);
+    locktidTrackerMutex();
```
dlherms-ibm: We used to lock for very short duration around the copy of the pointer...  Now, the lock is held for a longer duration and performing operations that could take an exception.  Maybe add try catch and ensure that the mutex is unlocked...

utilities/src/tracksyscall.cc
```diff
@@ -0,0 +1,56 @@
...
+#include "tracksyscall.h"
+
+
+static pthread_mutex_t tidTrackerMutex = PTHREAD_MUTEX_INITIALIZER;
+std::map<pthread_t, TrackSyscallPtr>   pthread_syscalltracker;
+
+std::map<pthread_t, TrackSyscallPtr> get_pthread_syscalltracker(){return pthread_syscalltracker;}
+
+void TrackSyscall::reset()
+{
```
dlherms-ibm: Not all data members are reset here...  Not sure if this could be a problem or not...  Maybe a common routine to init on construct as well as for reset()...

bb/src/tracksyscallMonitor.cc
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- |    tracksyscall.h
+ |    tracksyscall
```
dlherms-ibm: Nit...  Should be tracksyscallMonitor.cc and prolog for utilities/src/tracksyscall should be entitled tracksyscall.cc

utilities/src/tracksyscall.cc
```diff
@@ -0,0 +1,56 @@
...
+    _timeStamp = 0;
+    _syscall = nosyscall;
+    _fd = -1;
+    _timeStamp = 0;
+    _lineNumber = 0;
+    _rasCount = 0;
+    _size=0;
+    _offset=0;
+    pthread_mutex_lock(&tidTrackerMutex);
+    pthread_syscalltracker[_tid] = this;
```
dlherms-ibm: The syscalltrackers used to be for a rather static set of threads.  Now with runCommand, threads will be coming and going.  Maybe I missed it, but not sure if these are ever removed from the map...  Maybe not a problem if we leave them in there...

utilities/src/tracksyscall.cc
```diff
@@ -0,0 +1,56 @@
...
+{
+    _syscall = nosyscall;
+    _timeStamp=0;
+    _lineNumber=0;
+    _tid=-1;
+    _fd=-1;
+    _fileName.clear();
+}
+
+TrackSyscall::TrackSyscall()
```
dlherms-ibm: Not sure how dangerous it is here, but not sure if you want to insert into the map as part of the constructor.  If that were to fail for some reason, you take an exception in the constructor (ugh).  Personally, I'd write a method to create the TrackSyscall object (only initializing data members) and then insert into the map.  You could catch any errors on the insert and also send a rc back to the invoker.  But, that might be just my preference...

utilities/src/tracksyscall.cc
meahoibm: Did clear of string _filename and removed duplicate _timestamp=0

utilities/src/tracksyscall.cc
meahoibm: Removed reset as not used anywhere.

bb/src/tracksyscallMonitor.cc
meahoibm: Fixed prologue nit.

utilities/src/tracksyscall.cc
meahoibm: Good find. I need to look at thread_local destruction.

utilities/src/tracksyscall.cc
meahoibm: Constructor is thread_local.  I need to think how to mitigate.  I did it because it was easy, but you make a good point.
thread_local TrackSyscallPtr threadLocalTrackSyscallPtr=new TrackSyscall();
Maybe I need to call a factory method or stand-alone method?

Issue #776, opened 2019-07-31
Basic LSF jobs failing with TERM_EXTERNAL_SIGNAL on latest CSM


tgooding:
LSF jobs started failing a couple days ago.  Tracked this down to CSM changes.  Based on the timeframe, I suspect commit a29b8a31e312760f85db2adfe2cf75b9b2d5e894 might be related.  

Basic hostname:
```
bash-4.2$ bsub -q bbq -nnodes 1 hostname
Job <194560> is submitted to queue <bbq>.
```

bhist output:
```
bash-4.2$ bhist -l 194560

Job <194560>, User <tgooding>, Project <default>, Command <hostname>
Tue Jul 30 21:43:38: Submitted from host <c650mnp03>, to Queue <bbq>, CWD </var
                     /log/ibm/csm>, Re-runnable, Requested Resources <1*{select
                     [LN]span[hosts=1]} + 20*{select[CN]span[ptile=20]}>;
Tue Jul 30 21:43:38: Dispatched 21 Task(s) on Host(s) <1*c650mnp03> <20*c650f06
                     p29>, Allocated 21 Slot(s) on Host(s) <1*c650mnp03> <20*c6
                     50f06p29>, Effective RES_REQ <1*{select[(LN) && (type == a
                     ny)] order[r15s:pg] span[hosts=1]}+20*{select[(CN) && (typ
                     e == any)] order[r15s:pg] span[ptile=20]}>;
Tue Jul 30 21:43:38: Starting (Pid 31647);
Tue Jul 30 21:43:41: Exited by LSF signal TERM_EXTERNAL_SIGNAL. The CPU time us
                     ed is 0.0 seconds;
Tue Jul 30 21:43:41: Completed <exit>; TERM_EXTERNAL_SIGNAL: job killed by a si
                     gnal external to LSF;


PENDING TIME DETAILS:
Eligible pending time (seconds):	0
Ineligible pending time (seconds):	0

Summary of time in seconds spent in various states by  Tue Jul 30 21:43:41
  PEND     PSUSP    RUN      USUSP    SSUSP    UNKWN    TOTAL
  0        0        3        0        0        0        3           

```

fpizzano:
Recreated on CSM dev cluster using:

```
[root@c650mnp06 ~]# rpm -qa | grep lsf
lsf-server-10.1-20190722.ppc64le
lsf-common-10.1-20190722.ppc64le
lsf-master-10.1-20190722.ppc64le
[...]

PR #782, opened 2019-08-06
Issue #776: Version wasn't checked in init code. Version now clamps.

mew2057:
Fixing a bug in the x-macro for initialization. Version is now checked init code.

### DIFF
--- a/csmi/include/struct_generator/c_funct.txt
+++ b/csmi/include/struct_generator/c_funct.txt
@@ -213,16 +213,19 @@ FREE_FUNCT(CSMI_STRUCT_NAME)
     uint32_t i = 0;             //< A counter variable.
     #endif
 
+    // Get the version the user is supplying.
+    uint64_t version_id    = target->_metadata <= CSM_VERSION_ID ? target->_metadata : CSM_MIN_VERSION;
+
     #define CSMI_STRUCT_MEMBER(type, name, serial_type, length_member, init_value, metadata) \
         FREE_ARRAY_##serial_type(type, name, length_member, metadata)
-    #define CSMI_VERSION_START(version)
-    #define CSMI_VERSION_END(hash)
+    #define CSMI_VERSION_START(version) if (version_id >= version ) {
+    #define CSMI_VERSION_END(hash) } 
     #include STRUCT_DEF
 
     #define CSMI_STRUCT_MEMBER(type, name, serial_type, length_member, init_value, metadata) \
         FREE_##serial_type(type, name, length_member, metadata)
-    #define CSMI_VERSION_START(version)
-    #define CSMI_VERSION_END(hash)
+    #define CSMI_VERSION_START(version) if (version_id >= version ) {
+    #define CSMI_VERSION_END(hash) } 
     #include STRUCT_DEF
 }
 
--- a/csmi/include/struct_generator/c_funct.txt
+++ b/csmi/include/struct_generator/c_funct.txt
@@ -230,11 +230,15 @@ INIT_FUNCT(CSMI_STRUCT_NAME)
 {
     if (!target) return; //< If the target is null return. TODO Error throwing.
     
+    // Get the version the user is supplying.
+    uint64_t version_id    = target->_metadata <= CSM_VERSION_ID ? target->_metadata : CSM_MIN_VERSION;
+
+    
     // Initialize the members.
     #define CSMI_STRUCT_MEMBER(type, name, serial_type, length_member, init_value, metadata)\
         INIT_##serial_type(name, init_value, length_member)
-    #define CSMI_VERSION_START(version)
-    #define CSMI_VERSION_END(hash)
+    #define CSMI_VERSION_START(version) if (version_id >= version ) {
+    #define CSMI_VERSION_END(hash) } 
     #include STRUCT_DEF
 }
 #undef STRUCT_DEF
--- a/csmi/src/bb/src/csmi_bb_serialization.c
+++ b/csmi/src/bb/src/csmi_bb_serialization.c
@@ -250,11 +250,15 @@ INIT_FUNCT(CSMI_STRUCT_NAME)
 {
     if (!target) return; //< If the target is null return. TODO Error throwing.
     
+    // Get the version the user is supplying.
+    uint64_t version_id    = target->_metadata <= CSM_VERSION_ID ? target->_metadata : CSM_MIN_VERSION;
+
+    
     // Initialize the members.
     #define CSMI_STRUCT_MEMBER(type, name, serial_type, length_member, init_value, metadata)\
         INIT_##serial_type(name, init_value, length_member)
-    #define CSMI_VERSION_START(version)
-    #define CSMI_VERSION_END(hash)
+    #define CSMI_VERSION_START(version) if (version_id >= version ) {
+    #define CSMI_VERSION_END(hash) } 
     #include STRUCT_DEF
 }
 #undef STRUCT_DEF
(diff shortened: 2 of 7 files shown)


csmi/include/struct_generator/c_funct.txt
```diff
@@ -213,28 +213,35 @@ FREE_FUNCT(CSMI_STRUCT_NAME)
     uint32_t i = 0;             //< A counter variable.
     #endif
 
+    // Get the version the user is supplying.
+    uint64_t version_id    = target->_metadata <= CSM_VERSION_ID ? target->_metadata : CSM_MIN_VERSION;
```
besawn: I spent a lot of time reviewing this PR and the general version handling code, and I don't think this line is quite what I would expect. I think there are two cases to consider here:
A CSM client built against an older version of CSM talking to a newer version of CSM and the reverse case.

Consider two CSM versions: CSM_VERSION_1_6_0 and CSM_VERSION_1_7_0
When a client built against CSM_VERSION_1_6_0 connects to a daemon built against CSM_VERSION_1_7_0, code will evaluate like this:
```
uint64_t version_id = target->_metadata <= CSM_VERSION_ID ? target->_metadata : CSM_MIN_VERSION;
uint64_t version_id = CSM_VERSION_1_6_0 <= CSM_VERSION_1_7_0 ? CSM_VERSION_1_6_0 : CSM_MIN_VERSION;
uint64_t version_id = CSM_VERSION_1_6_0;
```
This looks like what I would expect in this case.

When a client built against CSM_VERSION_1_7_0 connects to a daemon built against CSM_VERSION_1_6_0, code will evaluate like this:
```
uint64_t version_id = target->_metadata <= CSM_VERSION_ID ? target->_metadata : CSM_MIN_VERSION;
uint64_t version_id = CSM_VERSION_1_7_0 <= CSM_VERSION_1_6_0 ? CSM_VERSION_1_7_0 : CSM_MIN_VERSION;
uint64_t version_id = CSM_MIN_VERSION;
uint64_t version_id = CSM_VERSION_0_4_1;
```
This looks wrong to me. When a newer client connects to an older CSM, we clamp to CSM_VERSION_0_4_1, but I think the daemon should be able to handle the client clamping to the same version as the daemon, CSM_VERSION_1_6_0.

I think re-writing this line as:
`uint64_t version_id    = target->_metadata <= CSM_VERSION_ID ? target->_metadata : CSM_VERSION_ID;`
would clamp to the current daemon version when a newer client connects.

csmi/include/struct_generator/c_funct.txt
```diff
@@ -213,28 +213,35 @@ FREE_FUNCT(CSMI_STRUCT_NAME)
...
+    #define CSMI_VERSION_END(hash) } 
     #include STRUCT_DEF
 }
 
 INIT_FUNCT(CSMI_STRUCT_NAME)
 {
     if (!target) return; //< If the target is null return. TODO Error throwing.
     
+    // Get the version the user is supplying.
+    uint64_t version_id    = target->_metadata <= CSM_VERSION_ID ? target->_metadata : CSM_MIN_VERSION;
```
besawn: Same comment about clamping to CSM_VERSION_ID instead of CSM_MIN_VERSION.

besawn:
Two additional comments:
I spent a lot of time reviewing, but I think some more in depth testing will reveal more than code reviews and vanilla regression test in this case. Did you attempt to test these changes with mixed CSM client and daemon versions?

If not, one suggestion for a possibly easy set of tests:
Install CSM 1.6 GA version and copy the contents of /opt/ibm/csm/bin to a temporary directory.
Uninstall CSM 1.6 and instead install the version of CSM built against this pull request.
Once the new CSM is installed, replace all of the files in /opt/ibm/csm/bin with the old versions and then run a full regression.

Then repeat the process above but with the CSM API command lines from this PR running against CSM 1.6 GA version daemons.

Final thing I noticed, this is more for future support releases, nothing currently needs to be done with respect to this comment:
We set the version ID based on number of commits in the git branch, I think:
```
#define CSM_VERSION_1_5_0 66816
#define CSM_VERSION_1_4_0 66560
#define CSM_VERSION_1_3_0 66304
#define CSM_VERSION_1_1_0 65792
#define CSM_VERSION_1_0_0 65536
#define CSM_VERSION_0_4_1 1025
```
So I think it is not safe for us to create a new CSM version in a efix branch, for example, if we decided to create a CSM_VERSION_1_6_1 in the cast_1_6_x branch, it could be that the version number of CSM_VERSION_1_6_1 ends up being larger than CSM_VERSION_1_7_0, depending on how many commits were made in cast_1_6_x vs master after the split between the branches. Just something I wasn't aware of that we will have to watch out for in the future.

csmi/include/struct_generator/c_funct.txt
mew2057: That doesn't account for the case where metadata is unset and has an arbitrarily high number in memory. We don't support forward compatibility for cumulative fixes in CSM. We do support forward compatibility for e-fixes, but those are generally more special cases.

Additionally, this is not an allocator. The `init` function just sets CSM defaults (that's why it requires a struct to be supplied). 

`CSM_MIN_VERSION` is guaranteed to be *safe* in these functions.  `CSM_VERSION_ID` in a bad case (e.g. metadata not being set) will result in the same problem. 

Really neither `init`, nor `free` should not be called directly, but that's a separate issue.

mew2057:
> Did you attempt to test these changes with mixed CSM client and daemon versions?
No, but this is a good candidate for regression. It had been on my queue for sometime, but it might be better for another team member to write this regression.

> We set the version ID based on number of commits in the git branch, I think:

No, the version id is calculated in the `regenerate_headers.sh` script. Specifically in the `regenerate_headers_functions.sh` script:
``` bash
 # Numeric constants for versioning.
 VERSION_WIDTH=8       # The width of a version code (major, minor and ptf have the same width).
 VERSION_MASK=$(( (1<<${VERSION_WIDTH}) - 1))

 # The verision shifts
 PTF_VERSION_SHIFT=0
 MINOR_VERSION_SHIFT=${VERSION_WIDTH}
 MAJOR_VERSION_SHIFT=$(( ${VERSION_WIDTH} + ${MINOR_VERSION_SHIFT} ))
```
```bash
active_version_num=$(( (${ptf_vers} << ${PTF_VERSION_SHIFT}) \
         + (${minor_vers} << ${MINOR_VERSION_SHIFT})              \
         + (${major_vers} << ${MAJOR_VERSION_SHIFT}) ))
```

This versioning schema allocates 256 version levels for every number in the version code.

csmi/include/struct_generator/c_funct.txt
mew2057: Additionally, the error that this is directly referencing may in fact hit the segmentation fault since they aren't using the initialize macro I wrote.

To really future poof code, `init` nor `free` should be used in API user code. They couple your code too closely to the version. It's better for the user to fully own their input structs.

besawn: > That doesn't account for the case where metadata is unset and has an arbitrarily high number in memory.
If it is possible for the metadata to be unset and thus initialized to garbage, this line is broken in all situations, not just the case where it happens to be arbitrarily high garbage.

mew2057: Fair, but I opted for the safest fallback in the case of arbitrarily high garbage. I don't mind if `CSM_MIN_VERSION` or `CSM_VERSION_ID` gets used.

besawn:
@mew2057 Thanks for the clarification on how the version number is calculated.

PR #818, opened 2019-11-08
Csm elk 6 8 1 compatibility

NickyDaB:
# Purpose
Provide CSM nov release with comatibility for ELK 6.8.1

# Approach
Discussed with team members and long hours of trial and error testing.


NickyDaB:
merging

PR #819, opened 2019-11-11
BSCFS_start fork change to work in tcsh (csh) and bash shells

meahoibm:
BSCFS_start change to run the agent in csh or bash remotely by ssh and set the job env correctly.


meahoibm:
Ran from bash as meaho on c650mnp03 to remote csh on p29-p33 with robot:
```[1756][meaho@c650mnp03:~]
$ robot -s bscfs /opt/ibm/bb ```

PR #883, opened 2020-04-08
Export layout leases debug help

meahoibm:
Changes to flag lease/close window.


tgooding:
Change looks good to me.

PR #884, opened 2020-04-09
Csm rh8 build

NickyDaB:
# Purpose
this will help get csm building on rh8

# Origin
- https://github.ibm.com/CSM/team_process/issues/69
- https://github.ibm.com/CSM/team_process/issues/67
- https://github.ibm.com/CSM/team_process/issues/63

# How to Test
- we want to try to build another rpm after this has been pulled in.
- set up a csm env
- run an allocation
- basically the same test nick has already run on his own branch
- then confirming they all work as expected, give nate and IST the thumbs up to use

## eventual TODOs
- python not supported
- dcgm not supported
- but enough working to get started (ie you can make allocations with cgroups and such, run apis, install the database, etc etc)


csmd/src/daemon/tests/csm_nodeset_test.cc
```diff
@@ -516,8 +517,8 @@ int MTCPerformanceTest( const int listSize, const unsigned mtcSize )
 
   size_t overlappedOffset = listSize/2;
   size_t overlappedNodes = overlappedOffset * (AGG_COUNT-2);
-  double minimal = ceil( (double)mtcSize / ((double)listSize/2) ) + 1;
-  double maximal = ceil( (double)mtcSize / ((double)listSize/2) ) + 2;
+  double minimal = std::ceil( (double)mtcSize / ((double)listSize/2) ) + 1;
```
tgooding, line 520: std::ceil is very different overhead/performance from ceil().  Why was this change needed?

csmd/src/daemon/src/csm_environmental_data.cc
```diff
@@ -164,7 +164,7 @@ void CSM_Environmental_Data::CollectNodeData()
   info = localtime( &rawtime );
 
   strftime(time_stamp_buffer, 80, "%Y-%m-%d %H:%M:%S", info);
-  snprintf(time_stamp_with_usec, 80, "%s.%06lu", time_stamp_buffer, now_tv.tv_usec);
+  snprintf(time_stamp_with_usec, 101, "%s.%06lu", time_stamp_buffer, now_tv.tv_usec);
```
lasch, line 167: When bumping up the allowed number of characters for snprintf, you need to increase the size of the corresponding buffer too! Otherwise you're at risk for out-of-bounds access. Line 157 is where that buffer is created (currently with size 80).

csmnet/src/CPP/endpoint_unix.cc
```diff
@@ -269,7 +269,12 @@ csm::network::EndpointUnix::~EndpointUnix()
 
   csm::network::AddressUnix_sptr addr = std::dynamic_pointer_cast<csm::network::AddressUnix>( GetLocalAddr() );
   if( addr == nullptr )
-    throw csm::network::Exception("Wrong address type", EBADF );
```
NickyDaB, line 272: @besawn @lasch 

Should we replace this with "return -1"?

lasch: I concur to remove the exception. How about this:
```
diff --git a/csmnet/src/CPP/endpoint_unix.cc b/csmnet/src/CPP/endpoint_unix.cc
index 23a230e..60f801e 100644
--- a/csmnet/src/CPP/endpoint_unix.cc
+++ b/csmnet/src/CPP/endpoint_unix.cc
@@ -269,9 +269,14 @@ csm::network::EndpointUnix::~EndpointUnix()
 
   csm::network::AddressUnix_sptr addr = std::dynamic_pointer_cast<csm::network::AddressUnix>( GetLocalAddr() );
   if( addr == nullptr )
-    throw csm::network::Exception("Wrong address type", EBADF );
-  if( addr->_SockAddr.sun_path[0] != 0 )
-    unlink( addr->_SockAddr.sun_path );
+  {
+    LOG(csmnet, critical ) << "EndpointUnix with wrong address type in destructor";
+  }
+  else
+  {
+    if( addr->_SockAddr.sun_path[0] != 0 )
+      unlink( addr->_SockAddr.sun_path );
+  }
 }
 
 int csm::network::EndpointUnix::Connect( const csm::network::Address_sptr aSrvAddr )
```
This would make sure the parent class destructor is called too. On the other hand, this check is mostly insurance against a situation that would have caused a ton of trouble before getting to the destructor (I think, I took care of those back then - except for remaining unspotted bugs). Thoughts?

NickyDaB:
@besawn and I met up, discussed, and debugged the python, boost, and cmake issues. 

We believe them to be in a much better state to move forward with. We believe that the original issues we had were due to a bad environment I was building in, which then had cascading effects. 

@tgooding you can view the current state of things, and re submit your review if you want. @besawn and I are meeting again monday morning to discuss other issues and next steps for completing the merge of this pull request.

csmd/src/daemon/src/csm_environmental_data.cc
besawn: I agree with Lars, we need to make a modification to this change.

Original compiler warning was:
```
csmd/src/daemon/src/csm_environmental_data.cc:167:38: error: \u2018%06lu\u2019 directive output may be truncated writing between 6 and 20 bytes into a region of size between 0 and 79 [-Werror=format-truncation=]
   snprintf(time_stamp_with_usec, 80, "%s.%06lu", time_stamp_buffer, now_tv.tv_usec);
                                      ^~~~~~~~~~
csmd/src/daemon/src/csm_environmental_data.cc:167:38: note: using the range [0, 18446744073709551615] for directive argument
/csmd/src/daemon/src/csm_environmental_data.cc:167:11: note: \u2018snprintf\u2019 output between 8 and 101 bytes into a destination of size 80
   snprintf(time_stamp_with_usec, 80, "%s.%06lu", time_stamp_buffer, now_tv.tv_usec);
```
The root cause of this warning is that `time_stamp_buffer` and `time_stamp_with_usec` are both the same size, but `time_stamp_with_usec` is intended to be everything in `time_stamp_buffer` plus the additional chars needed to add the usecs.

However, as Lars pointed out, it is not safe to increase the size of the write into the array without also increasing the size of the array. I think making the two structures different sizes and using a constant to represent the size of each array will make the situation safer and more clear.

Let's change this to:
```
  // Set _timestamp
  const int32_t TS_BUFF_SIZE(80);
  const int32_t TS_USEC_SIZE(120);
  char time_stamp_buffer[TS_BUFF_SIZE];
  char time_stamp_with_usec[TS_USEC_SIZE];

  struct timeval now_tv;
  time_t rawtime;
  struct tm *info;

  gettimeofday(&now_tv, NULL);
  rawtime = now_tv.tv_sec;
  info = localtime( &rawtime );

  strftime(time_stamp_buffer, TS_BUFF_SIZE, "%Y-%m-%d %H:%M:%S", info);
  snprintf(time_stamp_with_usec, TS_USEC_SIZE, "%s.%06lu", time_stamp_buffer, now_tv.tv_usec);
  _timestamp = time_stamp_with_usec;
```

csmd/src/daemon/tests/csm_daemon_echo_test.c
```diff
@@ -42,7 +42,8 @@ int client()
 
   csm_net_msg_t *msg;
   char *data = (char*)calloc(1024, sizeof(char));
-  strncpy( data, "Hello World", 10 );
+  //strncpy( data, "Hello World", 10 );
+  memcpy(data, "Hello World", 10 );
```
besawn, line 46: Original error was:
```
csmd/src/daemon/tests/csm_daemon_echo_test.c: In function \u2018client\u2019:
csmd/src/daemon/tests/csm_daemon_echo_test.c:45:3: error: \u2018strncpy\u2019 output truncated copying 10 bytes from a string of length 11 [-Werror=stringop-truncation]
   strncpy( data, "Hello World", 10 );
```
I think this should be fixed like this:
```
  strncpy( data, "Hello World", 12);

  msg = csm_net_msg_Init( CSM_CMD_ECHO, 1, CSM_PRIORITY_DEFAULT, 1243567, geteuid(), getegid(), data, 12, 0 );
```
I think it would be even better to fix it without the hard-coded `12` (which is the length of "Hello World" plus a null char), but I couldn't find a quick solution that did not seem like overkill.

csmd/src/daemon/src/csmi_request_handler/CSMIAllocationAgentUpdateState.cc
```diff
@@ -121,12 +121,19 @@ bool AllocationAgentUpdateState::HandleNetworkMessage(
...
                 "; Secondary Job Id: " << allocation->secondary_job_id <<
                 "; Error Code: "       << response->error_code << 
                 "; Error Message: "    << response->error_message << 
+                "; Message: Agent completed;";*/
+            LOG(csmapi,info) << ctx->GetCommandName() << ctx <<
+                "Allocation ID: "      << allocation->allocation_id << 
+                "; Primary Job Id: "   << allocation->primary_job_id << 
+                "; Secondary Job Id: " << allocation->secondary_job_id <<
+                "; Error Code: "       << response->error_code << 
+                "; Error Message: NICK WAS HERE"     <<
```
besawn, line 136: Please update this with a check to only print the value of `response->error_code` if it is not NULL.

csmd/src/daemon/tests/csm_nodeset_test.cc
besawn: @tgooding Can you expand on your statement about overhead/performance being different?

A change is needed due to the following compiler error when building on RHEL 8:
```
csmd/src/daemon/tests/csm_nodeset_test.cc: In function \u2018int NodeMTCTest(const NodeNameList_t&, const BitFilter_t*, unsigned int)\u2019:
csmd/src/daemon/tests/csm_nodeset_test.cc:195:20: error: \u2018ceil\u2019 was not declared in this scope
   double optimal = ceil( (double)mtcSize / ((double)aggSets[0].GetEntries()/2) ) + 2;
```

Simply adding math.h in the includes at the top of the file is sufficient to resolve the compiler error:
`#include <math.h>`

However, my understanding is that `cmath` is intended to be the C++ replacement library for `math.h`, so I would not expect there to be significant differences either way when choosing between `ceil` and `std::ceil` in this context.

NickyDaB: new version of file for review. 

https://github.com/IBM/CAST/pull/884/commits/dbd4ebf7447b9cccc0ccfb32c0bf2d4c03c2fc24

tgooding: I doubt you'd see a difference as you aren't math intensive in CSM.  However, ceil and std::ceil are different code bases with different branch/instruction pipelining, and std::ceil can conditionally throw exceptions.  I suspect you don't want that behavior and it probably needs additional plumbing in the library to figure that out.  Also C math routines have probably been aggressively optimized for all major platforms over the years to hit benchmarks.  The std C++ library has probably received less attention.  If you don't need std::ceil, its probably safer to avoid the switch at this point.

csmd/src/daemon/src/csm_environmental_data.cc
NickyDaB: new version of file for review.

https://github.com/IBM/CAST/pull/884/commits/706a79e9c08e1244908670964c706d194fd07e2b and https://github.com/IBM/CAST/pull/884/commits/ab1863365de970e5755efbb0d01a88806718143d

...and related files. 

because of related files. we can expand this out to a single include def file in the future. i will make an issue of cast noting the suggestions.

csmnet/src/CPP/endpoint_unix.cc
NickyDaB: Since this was @lasch original code, I implemented his proposed solution. 

https://github.com/IBM/CAST/pull/884/commits/9134396a600e5a359eeae479a4686ec0145433b3

please review the new state of the code

csmd/src/daemon/tests/csm_daemon_echo_test.c
NickyDaB: review changes here: https://github.com/IBM/CAST/pull/884/commits/3a9ec461d7be11caa36549394c19571cd15d5d08

csmd/src/daemon/src/csmi_request_handler/CSMIAllocationAgentUpdateState.cc
NickyDaB: code has been resolved in: https://github.com/IBM/CAST/pull/884/commits/80defba42f5b3e82facaeae6d55d12d2554651ac

Please review. 

Testing an allocation:

How the new implementation looks: 

Note: If there is no error message. It will not print an error message. But I still print the `Error Code` of 0. 

```
[COMPUTE]2020-04-13 13:59:34.716619     csmapi::info     | csm_allocation_create[485682771]; Allocation ID: 24; Primary Job Id: 1; Secondary Job Id: 0; Error Code: 0; Message: Agent completed;
```

Everything looks good.

NickyDaB:
@besawn Updated everything based off our code reviews. We should be good to run the regression tests with @williammorrison2 now.

PR #897, opened 2020-04-28
ansible playbooks for burst buffer install

meahoibm:
Adding bb/ansible directory of playbooks included in a master playbook for updating burstbuffer installs.


tgooding:
Closing this pull request, as mentioned above - something seems wrong.  Think it may have been already pull in via other pull requests.

PR #898, opened 2020-04-28
Ansible bb Playbook for creating and placing nodelist file

meahoibm:
Creates /tmp/nodelist on localhost where playbooks is running and then places in /etc/ibm on launch node(s).


tgooding:
Please add the standard comment block at the beginning of the yml.

PR #900, opened 2020-05-03
Various fixes for RHEL8, robot, python3

tgooding:


bb/src/LVUtils.cc
```diff
@@ -247,7 +247,8 @@ int doInitializeFileSystem(const char* pVolumeGroupName, const char* pDevName, c
                 char* l_DevName = new char[l_Length];
                 snprintf(l_DevName, l_Length, "/%s/%s/%s", DEVICE_DIRECTORY, pVolumeGroupName, pDevName);
                 char l_Cmd[1024] = {'\0'};
-                snprintf(l_Cmd, sizeof(l_Cmd), "mkfs.xfs -f -b log=%d %s 2>&1;", pLogBlockSize, l_DevName);
+//                snprintf(l_Cmd, sizeof(l_Cmd), "mkfs.xfs -f -b log=%d %s 2>&1;", pLogBlockSize, l_DevName);
+                snprintf(l_Cmd, sizeof(l_Cmd), "mkfs.xfs -f -b size=%d %s 2>&1;", 65536, l_DevName);  // \todo fixup
```
dlherms-ibm, line 251: Might consider just changing the constant in bbproxy.h...  The comment seems to be already correct...

const int DEFAULT_LOGBLOCKSIZE=16; //Default Block Size of 2**16 is 65536 bytes

bb/src/main.cc
```diff
@@ -90,8 +90,12 @@ uint64_t g_TimeBaseScale;
 
 uint64_t determineTimeBaseScale()
 {
-	uint64_t timebaseScale = 1;
-
+	static uint64_t timebaseScale = 0;
```
dlherms-ibm, line 93: Don't quite understand this change... The intent is to use g_TimeBaseScale as the global timebase scale factor on bbProxy and bbServer...

flightlog/src/flightlog.c
```diff
@@ -94,6 +94,57 @@ int FL_AttachRegistry(FlightRecorderRegistryList_t** reglist, const char* filena
 	return rc;
 }
 
+static uint64_t getTimeBaseScale()
+{
+	static uint64_t timebaseScale = 0;
```
dlherms-ibm, line 99: Similar comment as made in main.cc

bb/src/LVUtils.cc
tgooding: The todo comment was to cleanup the LOG references and switch it to bytes through the various function calls.  mkfs.xfs has removed the -b log=<num> syntax.

bb/src/main.cc
tgooding: We had roughly the same timebase parsing code in 3 places.  This change unifies 2 of those parsers (the 3rd is in flightlog).  Secondly, the 4.18 kernel is apparently 100x slower reading /proc/cpuinfo - not sure yet if this is a config thing on our system, or got a lot more expensive.  This accumulated causing 1.5 seconds of startup time alone and really impacted bbcmd.  The change should to read the value once.

PR #911, opened 2020-05-11
Csm ssl coms

NickyDaB:
# Purpose
With the move to RH8, SSL coms for csm daemons wasn't working. This pull request and code change fixes SSL coms for CSM daemons. 

# Origin
https://github.ibm.com/CSM/team_process/issues/71

# How to Test
testing is described in https://github.ibm.com/CSM/team_process/issues/71

# Screenshots
results of testing and code screenshots are in https://github.ibm.com/CSM/team_process/issues/71

### DIFF
--- a/csmnet/src/CPP/endpoint_ptp_sec.cc
+++ b/csmnet/src/CPP/endpoint_ptp_sec.cc
@@ -2,7 +2,7 @@
 
     csmnet/src/CPP/endpoint_ptp_sec.cc
 
-  © Copyright IBM Corporation 2015-2018. All Rights Reserved
+  © Copyright IBM Corporation 2015-2020. All Rights Reserved
 
     This program is licensed under the terms of the Eclipse Public License
     v1.0 as published by the Eclipse Foundation and available at
@@ -429,22 +429,36 @@ void csm::network::EndpointPTP_sec_base::SetupSSLContext( const csm::network::SS
 
   if( _LocalAddr->GetAddrType() == csm::network::AddressType::CSM_NETWORK_TYPE_AGGREGATOR )
   {
-    // We had "SSLv23_method" pre RH8. 
-    // Lars suggest that we change to TLS. I'm not sure why, but he knows more about it than me. 
-    // Leaving here incase it breaks again in the future and we have to debug around this. 
-    // This might be a note of different methods to investigate. 
-
-    //_gSSLContext = SSL_CTX_new( SSLv23_method() );
-    _gSSLContext = SSL_CTX_new( TLS_method() );
+    #ifdef TLS_method
+      // Use the recommended TLS_method if it is available (RHEL 8)
+      _gSSLContext = SSL_CTX_new( TLS_method() );
+    #else
+      // Otherwise fallback to the old behavior and use SSLv23_method (RHEL 7)
+      _gSSLContext = SSL_CTX_new( SSLv23_method() );
+    #endif
   }
   else
   {
     if( IsServerEndpoint() )
-      //_gSSLContext = SSL_CTX_new( SSLv23_server_method() );
-      _gSSLContext = SSL_CTX_new( TLS_server_method() );
+    {
+      #ifdef TLS_server_method
+        // Use the recommended TLS_server_method if it is available (RHEL 8)
+        _gSSLContext = SSL_CTX_new( TLS_server_method() );
+      #else
+        // Otherwise fallback to the old behavior and use SSLv23_server_method (RHEL 7)
+        _gSSLContext = SSL_CTX_new( SSLv23_server_method() );
+      #endif
+    }
     else
-      //_gSSLContext = SSL_CTX_new( SSLv23_client_method() );
-      _gSSLContext = SSL_CTX_new( TLS_client_method() );
+    {
+      #ifdef TLS_client_method
+        // Use the recommended TLS_client_method if it is available (RHEL 8)
+        _gSSLContext = SSL_CTX_new( TLS_client_method() );
+      #else
+        // Otherwise fallback to the old behavior and use SSLv23_client_method (RHEL 7)
+        _gSSLContext = SSL_CTX_new( SSLv23_client_method() );
+      #endif
+    }
   }
 
   if( _gSSLContext == nullptr )
--- a/csmnet/src/CPP/endpoint_ptp_sec.cc
+++ b/csmnet/src/CPP/endpoint_ptp_sec.cc
@@ -178,7 +178,17 @@ csm::network::EndpointPTP_sec_base::SendMsgWrapper( const struct msghdr *aMsg, c
     spos += aMsg->msg_iov[n].iov_len;
   }
 
-  int rc = BIO_write( _BIO, _SendBuffer, totalLen );
+  // RH 8 - CSM 1.8 - Lars and Nick fix
+
+  // Before RH8, we had a BIO write here. When we switched to RH8, SSL coms between daemons broke.
+  // we beleived that the SSL cert wasn't 100% verified or handeled correctly.
+  // By calling SSL_write() instead of BIO_write(), it seemed like the connection and verification was handled properly and cleanly.
+  // leaving this comment and original code here in case a future version creates a similar problem. 
+  // This may help guide a future support debug. 
+
+  //int rc = BIO_write( _BIO, _SendBuffer, totalLen );
+
+  int rc = SSL_write( _SSLStruct, _SendBuffer, totalLen );
 
   if( rc < 0 )
   {
@@ -193,7 +203,7 @@ csm::network::EndpointPTP_sec_base::SendMsgWrapper( const struct msghdr *aMsg, c
     }
   }
 
-  LOG(csmnet, debug) << "PTP::SendMsg: "
+  LOG(csmnet, debug) << "PTP_sec::SendMsg: "
       << " total_len=" << aMsg->msg_iov[0].iov_len + aMsg->msg_iov[1].iov_len
       << " rc=" << rc << " errno=" << errno;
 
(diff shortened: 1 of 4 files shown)


csmnet/src/CPP/endpoint_ptp_sec.cc
```diff
@@ -240,22 +250,105 @@ csm::network::EndpointPTP_sec_base::Recv( csm::network::Message &aMsg )
...
+            // logically this would be the same path as:
+            /*
+            if (BIO_should_retry( _BIO ) )
+              return 0;
+            */
+
+            // So I'll also return 0.
+            // But after my reading of the man page, I think there could be room for improvement here. 
+
+            return 0;
```
lasch, line 333: Just as a clarification: return 0 is the correct behavior for the endpoint::recv call because it means that currently there's not more data to receive so we shouldn't block on another recv call and instead see if any other endpoint has inbound data or any outbound work is to be done.

csmnet/src/CPP/endpoint_ptp_sec.cc
```diff
@@ -178,7 +178,17 @@ csm::network::EndpointPTP_sec_base::SendMsgWrapper( const struct msghdr *aMsg, c
...
+
+  // Before RH8, we had a BIO write here. When we switched to RH8, SSL coms between daemons broke.
+  // we beleived that the SSL cert wasn't 100% verified or handeled correctly.
+  // By calling SSL_write() instead of BIO_write(), it seemed like the connection and verification was handled properly and cleanly.
+  // leaving this comment and original code here in case a future version creates a similar problem. 
+  // This may help guide a future support debug. 
+
+  //int rc = BIO_write( _BIO, _SendBuffer, totalLen );
+
+  int rc = SSL_write( _SSLStruct, _SendBuffer, totalLen );
```
lasch, line 191: If I'm not mistaken, SSL_write() needs the same or similar error handling as SSL_read()

csmnet/src/CPP/endpoint_ptp_sec.cc
```diff
@@ -355,8 +456,16 @@ void csm::network::EndpointPTP_sec_base::SetupSSLContext( const csm::network::SS
...
+  // Check to see how to set up peer verification
+  // servers and clients need to verify differently. 
+  int verifyer_flags = SSL_VERIFY_PEER;
+  if( IsServerEndpoint() )
+  {
+    verifyer_flags |= SSL_VERIFY_CLIENT_ONCE | SSL_VERIFY_FAIL_IF_NO_PEER_CERT;
+  }  
+   SSL_CTX_set_verify( _gSSLContext, verifyer_flags, nullptr );
+   SSL_CTX_set_verify_depth( _gSSLContext, 1 );
+   SSL_CTX_set_mode( _gSSLContext, SSL_MODE_AUTO_RETRY );
```
lasch, line 468: AUTO_RETRY forces the SSL_read/write() to wait for actual user data to appear and forces any SSL mgmt traffic to finish.   If we keep the AUTO_RETRY here, we need to keep a close look on corner cases where the peer stops responding (e.g. due to network outages or cable pulls). This might cause the daemon to block and get stuck on this endpoint.
Since you have added the SSL error handling to SSL_read() (and should add it to SSL_write() too), I think we don't need the AUTO_RETRY any more.

csmnet/src/CPP/endpoint_ptp_sec.h
```diff
@@ -153,6 +155,48 @@ class EndpointPTP_sec_base : public EndpointPTP_base {
...
+    // Debug prints to help with SSL issues.
+    // Print out connection details
+    LOG( csmnet, debug ) << "SSL_get_version: " << SSL_get_version(newssl);
+    LOG( csmnet, debug ) << "SSL_get_cipher: " << SSL_get_cipher(newssl);
+
+    // The reference count of the X509 object is incremented by one, 
+    // so that it will not be destroyed when the session containing the peer certificate is freed. 
+    // The X509 object must be explicitly freed using X509_free().
+    X509_free(peer_cert);
+
```
lasch, line 199: There are a bunch of todo: cleanup XYZ in this section and before it. Do you want to address those in a separate PR/issue?

csmnet/src/CPP/endpoint_ptp_sec.cc
NickyDaB: ok. ill add that into the comment.

csmnet/src/CPP/endpoint_ptp_sec.cc
NickyDaB: I'll add in additional error handling

csmnet/src/CPP/endpoint_ptp_sec.cc
NickyDaB: ok. yeah i added the error handling. 

i'll remove the auto retry.

csmnet/src/CPP/endpoint_ptp_sec.h
NickyDaB: that was there in the code before. this pull request is just suppose to get the daemons back into a working SSL mode. 

that should be made into a new issue with a separate pull request to resolve those todos/ clean ups.

csmnet/src/CPP/endpoint_ptp_sec.cc
NickyDaB: so for this guy I was looking into it. 

https://www.openssl.org/docs/man1.1.1/man3/SSL_get_mode.html

removing the line just never sets it. but it doesnt necessarily start in that mode.

"SSL_CTX_set_mode() adds the mode set via bitmask in mode to ctx. Options already set before are not cleared. SSL_CTX_clear_mode() removes the mode set via bitmask in mode from ctx."

i double checked in the code. if somehow we get into a state with SSL_MODE_AUTO_RETRY if we dont specifically say set to blank than we could potentially be in that mode. 

here i found what mode we were in, then print, then set to auto retry, then print, then turn auto retry then print
```
  long currentBitmask = SSL_CTX_get_mode(_gSSLContext);
  LOG( csmnet, debug ) << "SSL_get_mode: currentBitmask=" << currentBitmask;
  SSL_CTX_set_mode( _gSSLContext, SSL_MODE_AUTO_RETRY );
  currentBitmask = SSL_CTX_get_mode(_gSSLContext);
  LOG( csmnet, debug ) << "SSL_get_mode: currentBitmask=" << currentBitmask;
  SSL_CTX_clear_mode( _gSSLContext, SSL_MODE_AUTO_RETRY );
  currentBitmask = SSL_CTX_get_mode(_gSSLContext);
  LOG( csmnet, debug ) << "SSL_get_mode: currentBitmask=" << currentBitmask;
```
```
[AGG]2020-05-12 13:15:24.855038     csmnet::debug    | SSL_get_mode: currentBitmask=4
[AGG]2020-05-12 13:15:24.855077     csmnet::debug    | SSL_get_mode: currentBitmask=4
[AGG]2020-05-12 13:15:24.855107     csmnet::debug    | SSL_get_mode: currentBitmask=0
```

based off of step two, when i turn it on and it says bitmask 4. then i assume bitmask 4 means on.

so we start on, i turn it on, which does nothing because its already on, then i turn it off and it goes to 0.

I ran the test a couple of times. and if i interpret the man page correctly. then auto retry is always on by the time we get to this part of the code. (maybe by default, or some setting turns it on b4 here) 

if we want it off, then i can call `  SSL_CTX_clear_mode( _gSSLContext, SSL_MODE_AUTO_RETRY );`

i also tested both with it on and off.... i didnt see a difference in the output of the apis or in the logs. 

So based off this new information does that change your suggestion of turning off the auto retry?

or do you want me to add in the clear to ensure that auto retry is off . 

looking for @lasch  and @besawn  's  opinions. thanks.

NickyDaB:
@lasch  as for the "gen remark"

> General remark: with the change from BIO_read/write to SSL_read/write, we might have to review the use/creation of all the BIO crap. Some of it might have become obsolete now.

I'll open an issue for it as to not forget about it.

csmnet/src/CPP/endpoint_ptp_sec.cc
lasch: I don't expect us to see the effect of this setting in the simple test cases here. I guess, that would probably require cable-pull tests or something to see the blocking of recvs/sends with the AUTO_RETRY enabled. Also interesting to see that auto-retry is on by default. Well, we could as well keep this as is and keep a sticky note to the corner of the monitor saying: "If we start seeing stuck daemons with SSL enabled, it's probably the auto-retry in combination with ssl_read/write."

besawn:
These changes do not build on RHEL 7:
```
[  2%] Building CXX object csmnet/src/CMakeFiles/csm_network_cc_objs.dir/CPP/endpoint_ptp_sec.cc.o
/u/besawn/github/CAST.nick.RHEL7/csmnet/src/CPP/endpoint_ptp_sec.cc: In member function \u2018void csm::network::EndpointPTP_sec_base::SetupSSLContext(const csm::network::SSLFilesCollection&)\u2019:
/u/besawn/github/CAST.nick.RHEL7/csmnet/src/CPP/endpoint_ptp_sec.cc:438:44: error: \u2018TLS_method\u2019 was not declared in this scope
     _gSSLContext = SSL_CTX_new( TLS_method() );
                                            ^
/u/besawn/github/CAST.nick.RHEL7/csmnet/src/CPP/endpoint_ptp_sec.cc:444:53: error: \u2018TLS_server_method\u2019 was not declared in this scope
       _gSSLContext = SSL_CTX_new( TLS_server_method() );
                                                     ^
/u/besawn/github/CAST.nick.RHEL7/csmnet/src/CPP/endpoint_ptp_sec.cc:447:53: error: \u2018TLS_client_method\u2019 was not declared in this scope
       _gSSLContext = SSL_CTX_new( TLS_client_method() );
                                                     ^
make[2]: *** [csmnet/src/CMakeFiles/csm_network_cc_objs.dir/CPP/endpoint_ptp_sec.cc.o] Error 1
make[1]: *** [csmnet/src/CMakeFiles/csm_network_cc_objs.dir/all] Error 2
make: *** [all] Error 2
*** CONFIGURE FAILED  (cmd= make noinstall package   rc=512)
```
I'll investigate further...

besawn:
I have checked in a proposed fix to handle the build failures on RHEL 7:
https://github.com/IBM/CAST/pull/911/commits/68564d15a6dc9a6b41dbb1e86b69c154e0b01718

@lasch Tagging you just in case you are curious and want to take a look...

PR #913, opened 2020-05-12
Clear and regenerate csm db and then put nodes into service

meahoibm:
Stand-alone playbook for review.  Eventually will do as an import within update.yml


meahoibm:
Ready for review.

tgooding:
I'm not sure this sequence works.  I'm thinking wiping/reiniting of the database needs to be when CSM is not running.  And then once CSM is up and running (i.e., compute nodes have been discovered), you can marking the nodes as IN_SERVICE.

meahoibm:
Please review and let me know if I need to change anything.

besawn:
@tgooding and @meahoibm This playbook looks very useful for test automation purposes and extremely dangerous if run accidentally on a production system. How are you planning to package these changes up? Are you taking any precautions to prevent them from accidentally ending up installed on a production system?

tgooding:
Agree @besawn : maybe move the FVT automation pieces into a separate RPM that has a narrower distribution?

tgooding:
Is there a CMakeList to install files into burstbuffer-fvt-ansible?  Seems like a new part didn't get checked in.

PR #917, opened 2020-05-20
Fix rmmod export_layout OOPs when in use by bbproxy

meahoibm:
export_layout is now marked in use when /dev/export_layout is opened.  If export_layout was removed after an open against /dev/export_layout, the close (release) would OOPs.  Other OOPs situations were possible.  The use count is incremented on open and decrement on release.

bbproxy was changed to do a dummy action of open export_layout early to seize it so it could not be removed.  If /dev/export_layout could not be opened, bbproxy quits on the early discovery.

### DIFF
--- a/bb/src/bbproxy.cc
+++ b/bb/src/bbproxy.cc
@@ -4891,6 +4891,18 @@ int bb_main(std::string who)
     /* Perform any clean-up of existing logical volumes */
     logicalVolumeCleanup();
 
+    /* Seize /dev/export_layout so rmmod fails on in-use while daemon is running */
+    int inuse_fd = -1;
+    if(config.get(process_whoami+".use_export_layout", false)){
+        inuse_fd=open("/dev/export_layout", O_RDWR | O_CLOEXEC, 0);
+        if (inuse_fd<0){
+            LOG(bb,always) << "/dev/export_layout failure errno=" << errno;
+            throw runtime_error(string("Check export_layout module installation.  Open failed with errno=") + to_string(errno));
+        }
+          
+    }
+      
+
     rc = setupUnixConnections(process_whoami);
     if(rc)
     {
--- a/bb/src/bbproxy.cc
+++ b/bb/src/bbproxy.cc
@@ -4843,7 +4843,7 @@ int registerHandlers()
     return 0;
 }
 
-
+static  int inuse_fd = -1;
 char LVM_SUPPRESS[] = "LVM_SUPPRESS_FD_WARNINGS=1";  // note: ownership of this string is moved to ENV during putenv() call.
 int setupUnixConnections(string whoami);
 int bb_main(std::string who)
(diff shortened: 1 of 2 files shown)


bb/src/bbproxy.cc
```diff
@@ -4891,6 +4891,18 @@ int bb_main(std::string who)
     /* Perform any clean-up of existing logical volumes */
     logicalVolumeCleanup();
 
+    /* Seize /dev/export_layout so rmmod fails on in-use while daemon is running */
+    int inuse_fd = -1;
```
tgooding, line 4895: I suspect Coverity will flag this as a leaked file descriptor.  I'd suggest making this a global variable to avoid the complaint.  (and for good hygiene, close that file descriptor in the main() exit flow)

export_layout/src/export_layout.c
```diff
@@ -235,6 +235,7 @@ static int export_layout_open(struct inode *inodep, struct file *filep)
 	if (!t)
 		return -ENOMEM;
 	filep->private_data = t;
+        __module_get(THIS_MODULE);
```
tgooding, line 238: The double-underscore worries me, implies deprecated/internal that may go away in future kernel levels.  It looks like there is a try_module_get() variant that is similar - - is that more appropriate?

meahoibm: I will change to the try_module_get to avoid the module_get going away in the future.  
Here is the additional research I did.

The try_module_get is for a module looking to lock down a module to be used.  The safety is a check that the module is loaded and then incrementing the use count.  Since export_layout is upping the use count for itself, the pointer check is not necessary.  

The count of __module_get(THIS_MODULE) in the latest kernel git is 76.  So a lot of use.

The kernel doc lends support to your statement.  The Documentation/kernel-hacking/hacking.rst 
_:c:func:`try_module_get()`/:c:func:`module_put()`
479 -------------------------------------------------
480 
481 Defined in ``include/linux/module.h``
482 
483 These manipulate the module usage count, to protect against removal (a
484 module also can't be removed if another module uses one of its exported
485 symbols: see below). Before calling into module code, you should call
486 :c:func:`try_module_get()` on that module: if it fails, then the
487 module is being removed and you should act as if it wasn't there.
488 Otherwise, you can safely enter the module, and call
489 :c:func:`module_put()` when you're finished.
490 
491 Most registerable structures have an owner field, such as in the
492 :c:type:`struct file_operations <file_operations>` structure.
493 Set this field to the macro ``THIS_MODULE``._

bb/src/bbproxy.cc
meahoibm: Did the clean use of making global and closing on exit.  Please run coverity to check.

bb/src/bbproxy.cc
```diff
@@ -4925,6 +4924,7 @@ int bb_exit(std::string who)
     ENTRY_NO_CLOCK(__FILE__,__FUNCTION__);
 
 //    printf("bb_exit: who = %s\n", who.c_str());
+    close(inuse_fd);
```
tgooding, line 4927: Probably should be:  if(inuse_fd >= 0) close(inuse_fd);

PR #919, opened 2020-05-26
null terminate char buffer for coverity 338064

meahoibm:


meahoibm:
Since this a small change, should the merge conflicts be avoided by just redoing this push?

meahoibm:
This has a better fix in later code now in master.  Closing out.

Issue #918, opened 2020-05-20
Data loss when transferring sparse files between SSD and GPFS


tonyhutter:
**Describe the bug**
If you have a sparse file, or a partially sparse file (with the sparse sections after the extents), then the file will not be transferred correctly using the BBAPI between the SSD and GPFS.  If you transfer a file from the SSD to GPFS that is completely sparse:
```
$ truncate -s 1M file1
```
You'll get a 0 byte file in GPFS.  If you transfer a file that's partially sparse:
```
$ echo "hello world" > file2 && truncate -s 1M file2
```
You'll get a file that's 65536 bytes in GPFS.  

Full details:
https://github.com/LLNL/scr/issues/163#issuecomment-631734186

**To Reproduce**
Replace the `randfile` command in  https://github.com/IBM/CAST/blob/master/bb/tests/src/test_basic_xfer.c#L233 with one of the commands above.

**Expected behavior**
I expect the destination files to be the same size as the source files (1MB).

**Environment (please complete the following information):**
 - Machine [lassen, LLNL]
 - Version [ibm-csm-1.6.0-2603, ibm-burstbuffer-1.6.2.3135]

**Additional context**
https://github.com/LLNL/scr/issues/163#issuecomment-631734186

**Issue Source:**
https://github.com/LLNL/scr/issues/163#issuecomment-631734186

tgooding:
Looks like the apparent file size is incorrect but i'd surmise that all sparse data was transferred to GPFS intact.  BB looks at the allocated blocks and only transfers those.  However, if the file's last allocated block != last file block, then the file size (as reported by `ls`) won't match.  I think thats what you are seeing.  I'd agree that we should try to match the file metadata.  

@dlherms-ibm : can you look?

dlherms-ibm:
Will do...

tonyhutter:
Some additional stat and hexdump info:

```bash
$ echo "hello world" > hello
$ truncate -s 1M hello

$ ls -l hello 
-rw------- 1 hutter2 hutter2 1048576 May 21 09:08 hello

$ stat hello
  File: 'hello'
  Size: 1048576   	Blocks: 128        IO Block: 65536  regular file
Device: fd02h/64770d	Inode: 1027        Links: 1
Access: (0600/-rw-------)  Uid: (57592/ hutter2)   Gid: (57592/ hutter2)
Access: 2020-05-21 09:08:26.970249503 -0700
Modify: 2020-05-21 09:08:31.410053045 -0700
[...]

PR #921, opened 2020-05-27
bb:Fix for #918 Data loss when transferring sparse files between SSD …

dlherms-ibm:
…and GPFS

# Purpose
_Describe the problem or feature. Ask yourself, "What is the goal of this pull request?"_

# Approach
_How did you address the problem? What was your thinking process?_

# Origin
_Link the origin git issue here. and describe any offline discussion if applicable._

# How to Test
_This section will help the regression tester._
2. Test it.

# Screenshots
_Place appropriate screenshots or code snippets here._

_Make sure you attempted to do the following:_
- [ ] Appropriatly label
- [ ] Assign a milestone. 
- [ ] update appropriate documentation in the `/docs` folder for your code changes. 
- [ ] Remove Open Questions and Pre-Merge TODOs when you no longer need it.

### DIFF
--- a/bb/src/BBLV_ExtentInfo.cc
+++ b/bb/src/BBLV_ExtentInfo.cc
@@ -75,6 +75,7 @@ int BBLV_ExtentInfo::addExtents(const LVKey* pLVKey, const uint64_t pHandle, con
         vector<Extent> l_NewList;
         size_t l_StatSize = 0;
         size_t l_AccumulatedLength = 0;
+        size_t l_AccumulatedLength2 = 0;
         size_t l_LastExtentLength = 0;
         for (std::vector<Extent>::size_type i = 0; i < pTransferDef->extents.size(); ++i)
         {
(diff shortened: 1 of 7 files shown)


dlherms-ibm:
So, ftruncate was not supported via BBIO and I decided to not add it there.  Instead, I chose to use the existing support via filehandle directly.  We already create filehandles on the bbProxy side to simply house file stats.  This is the first case where we are doing this on the bbServer side.  As part of the second pass processing on bbServer for startTransfer, we were already calculating the total number of bytes to be transferred.  So, I decided to set a flag in bbflags indicating a sparse file if the bytes we are to transfer is less than the fstat size of the source file on the SSD.  Sparse files are only indicated for stageout and for normal files.  That indicator is what is used in removeFromInFlight() to perform a final ftruncate for the size of the source file.  All transfer processing goes through the removeFromInFlight() path, so the cleanup of the new filehandle should be caught in all cases.  I added new testcases in my environment for sparse files with a trailing hole.  Wasn't sure if we wanted to do the same in the robot bucket.

Code has been run successfully in two robot runs (no LSF).  Has run fine in my environments, including failover scenarios.

dlherms-ibm:
Working on marking the file/transfer as failed if the ftruncate fails...

Issue #925, opened 2020-06-03
csm_db data collection field "archive_history_time" is empty


thanh-lam:
**Describe the bug**
The CSM command /opt/ibm/csm/db/csm_db_history_archive.py stores resulted json files in /var/log/ibm/csm/archive/. The files csm_ib_cable_history-2020-22.json, csm_switch_history-2020-22.jso,  csm_switch_inventory_history-2020-22.json contain the field "archive_history_time" that's empty. That field shows up with the "null" value in the json files. And, it shows up as empty and a "warning mark" in Kibana Web UI.

Bill Morrison found "an issue related to the csm_db_history_archive.py script it self. We create a temp table in the DB that processes all the records with the achive_history_time as being null first. Once that is done we then update the actual history table with the archive_history_time stamp. The json string doesn't have the 'archive_history_time' in it. "
 
**To Reproduce**
1. Go to '/opt/ibm/csm/sbin'
2. Run the command './standalone_ib_and_switch_collection'
3. Go to '/opt/ibm/csm/db'
4. Run the command './csm_db_history_archive.py'
5. Go to '/var/log/ibm/csm/archive'
6. Run 'grep "archive_history_time" csm_db_ib* | head -1
7. Run 'grep "archive_history_time" csm_db_switch* | head -1

**Expected behavior**
The field "archive_history_time" should contain a date or timestamp when it's archived.

**Screenshots**

**Environment (please complete the following information):**
 - IST CSM BDS cluster: csm master: f4n05
 - CSM 1.8 build 3397

**Additional context**

**Issue Source:**
Testing CSM BDS updates to ELK 7.5.1 & CSM 1.8 on RHEL 8.1

williammorrison2:
### This is a very basic set of tests.
##### I added in this bit of the code to update both temp and history table in the `csm_db_history_archive.py` script.
```
    ## Update the flagged tables.
    cursor.execute("WITH m_update AS ( \
         UPDATE {0} SET archive_history_time  = 'now()' \
         FROM temp_{0} WHERE temp_{0}.id = {0}.ctid \
         AND {0}.archive_history_time IS NULL \
         AND {0}.{1} = temp_{0}.{1} \
         RETURNING * \
         ) \
         UPDATE temp_{0} \
         SET archive_history_time  = 'now()' \
         FROM m_update".format(table_name,time))
```

##### Some of the results from the `csm_ib_cable_history-2020-22.json` file
```
[...]

PR #933, opened 2020-06-16
This is a more efficient fix related to the achive history timestamp …

williammorrison2:
…issue #925

# Purpose
This fix will include the archive_history_time stamp in the json file when the csm_db_history_archive.py is executed

# Screenshots
Provided within issue #925


csmdb/sql/csm_db_history_archive.py
```diff
@@ -150,18 +150,21 @@ def dump_table( db, user, table_name, count, target_dir, is_ras=False ):
...
     (result,) = cursor.fetchall()
-    
+  
     ## Update the flagged tables.
-    cursor.execute("UPDATE {0} SET archive_history_time = 'now()' \
-        FROM temp_{0} WHERE temp_{0}.id = {0}.ctid \
-        AND {0}.archive_history_time IS NULL \
-        AND {0}.{1} = temp_{0}.{1}".format(table_name,time))
-    
+    cursor.execute("UPDATE {0} SET archive_history_time  = temp_{0}.archive_history_time \
```
besawn, line 163: I think it might possibly be more efficient to leave this as `'now()'`, but this is good, too.

PR #951, opened 2020-08-18
Adding in a hardcoded PostgreSQL INSERT statement that uses an active…

williammorrison2:
…  node in the cluster (eliminates an external sql script)

# Purpose
To use an environmental variable in a PostgreSQL INSERT, which can be done in other cluster testing.


williammorrison2:
Added in a logical step to check if the CSM inventory collection captures the SSD(s) on a single compute node.
1. If it does, then the FVT regression will test against that particular record in the DB.
2. If no SSD record is present from the single compute in the CSM DB, then we will insert a "dummy" record to use.

PR #973, opened 2020-09-22
Use dnf and potentially pull from a remote repo

tgooding:
Improve install playbooks

### DIFF
--- a/bb/ansible/bbInstall.yml
+++ b/bb/ansible/bbInstall.yml
@@ -12,6 +12,11 @@
 ###########################################################
 #  sudo ansible-playbook -i hosts  <this.yml>
 
+#check reachability of all hosts
+- hosts: all
+  tasks:
+  any_errors_fatal: true 
+
 - hosts: launch
   tasks:
   - name: install BB RPMs on launch node
--- a/bb/ansible/bbInstall.yml
+++ b/bb/ansible/bbInstall.yml
@@ -12,16 +12,10 @@
 ###########################################################
 #  sudo ansible-playbook -i hosts  <this.yml>
 
-- hosts: all
-  tasks:
-  - name: check all nodes for RPM directory
-    command: ls -d  {{BBRPMDIR}} 
-    any_errors_fatal: true
-
 - hosts: launch
   tasks:
   - name: install BB RPMs on launch node
-    shell: "yum localinstall -y {{BBRPMDIR}}/ibm-burstbuffer-[1tml]*rpm {{BBRPMDIR}}/ibm-flightlog-1*.rpm"
+    shell: "dnf install -y ibm-burstbuffer"
     args:
       warn: false # set warn=false to prevent warning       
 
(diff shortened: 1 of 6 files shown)


bb/ansible/bbInstall.yml
```diff
@@ -12,16 +12,10 @@
 ###########################################################
 #  sudo ansible-playbook -i hosts  <this.yml>
 
-- hosts: all
```
meahoibm, line 15: The purpose of the hosts: all step was twofold: check the RPMs are available AND that all the nodes are ready for processing.  Ansible will check all the hosts for reachability if it has not done so before.  The risk of not doing this step is that some steps will be executed and a reachability problem will show up later.  This could very well be OK since a restart of the playbooks as written are rather tolerant of incompleteness.

bb/ansible/bbInstall.yml
```diff
@@ -33,15 +27,12 @@
     ignore_errors: yes
 
   - name: Install BB RPMs on ESS
-    shell: "yum localinstall -y {{BBRPMDIR}}/ibm-burstbuffer-1*.rpm {{BBRPMDIR}}/ibm-flightlog-1*.rpm"
+    shell: "dnf install -y ibm-burstbuffer"
     args:
       warn: false # set warn=false to prevent warning
```
meahoibm, line 32: FYI:  Ansible did a warning previously for yum which required the warn: false since ansible treated yum as special.  Does dnf require this under ansible?  Probably harmless to do the warn: false still.

bb/ansible/bbInstall.yml
```diff
@@ -12,16 +12,10 @@
...
-  tasks:
-  - name: check all nodes for RPM directory
-    command: ls -d  {{BBRPMDIR}} 
-    any_errors_fatal: true
-
 - hosts: launch
   tasks:
   - name: install BB RPMs on launch node
-    shell: "yum localinstall -y {{BBRPMDIR}}/ibm-burstbuffer-[1tml]*rpm {{BBRPMDIR}}/ibm-flightlog-1*.rpm"
+    shell: "dnf install -y ibm-burstbuffer"
```
meahoibm, line 18: Does this mean accessing a repository for ibm-burstbuffer now?

bb/fvt-ansible/update.yml
```diff
@@ -25,21 +25,23 @@
 
     - debug: 
          msg: "cwd is {{pwd_info.stdout}} playbook_dir is {{playbook_dir}}" 
-    - debug:
-         msg: "BBRPMDIR={{BBRPMDIR}}"
-    - debug:
-        msg:  "CSMRPMDIR={{CSMRPMDIR}}"
```
meahoibm, line 31: Any useful debug information about dnf to put out?

bb/ansible/bbInstall.yml
tgooding: Yes.  Pointing to a local repo would need to have run "createrepo" in the work/rpms directory.

bb/ansible/bbInstall.yml
tgooding: Ah, ok.  I'll revert that piece of it.

meahoibm (approved):
Ready to roll.

Issue #981, opened 2020-10-08
Should BB API preserve permissions & xattrs to destination file


tonyhutter:
**Describe the bug**
Should the BB API preserve permissions and xattrs on the files it's transferring?  

I used your `test_basic_xfer` utility to copy a file on the burst buffer with `O777` permissions, and xattr "hello=world" set, to its destination on GPFS (with umask set to u=rwx,g=rwx,o=rwx).  After the transfer the destination file was `O750`, not `O777`, and the xattr was not set. 

**To Reproduce**
Make this change:
```diff

--- test_basic_xfer.c.orig	2020-05-20 15:26:40.926135000 -0700
+++ test_basic_xfer.c	2020-10-08 12:27:10.670129000 -0700
@@ -230,7 +230,7 @@
 
     if(dogenerate)
     {
-        snprintf(cmd, sizeof(cmd), "/opt/ibm/bb/tools/randfile --file=%s --by 512 --size=%ld", sfn, filesize);
+        snprintf(cmd, sizeof(cmd), "/opt/ibm/bb/tools/randfile --file=%s --by 512 --size=%ld && chmod 777 %s && attr -s hello -V world %s", sfn, filesize, sfn, sfn);
         if(rank == 0)
         {
             printf("Generate random file: %s\n", cmd);
```
Then:
```bash
# Check that destination file doesn't already exist
$ ls -l /p/gpfs1/hutter2/dst/rank.0
ls: cannot access /p/gpfs1/hutter2/dst/rank.0: No such file or directory

# Check umask
$ umask -S
u=rwx,g=rwx,o=rwx

# Do the transfer
$ ./test_basic_xfer 1 /p/gpfs1/hutter2/dst 1024
Source file: /tmp/bblv_hutter2_130450//rank.0
Target file: /p/gpfs1/hutter2/dst/rank.0
METADATA branch=(null)
METADATA gitcommit=(null)
METADATA job=(null)
METADATA type=out
METADATA jobid=130450
METADATA jobstep=2768240641
METADATA user=hutter2
METADATA pfspath=/p/gpfs1/hutter2/dst
METADATA bbpath=/tmp/bblv_hutter2_130450/
METADATA execname=./test_basic_xfer
METADATA filesize=1024
METADATA ranks=1
[...]

PR #984, opened 2020-10-27
Issue #981

meahoibm:
Process umask was unmasking bits on SSD to GPFS file copy on create in GPFS.
Reproduced problem.
Unit tested change in fh.cc


meahoibm:
@dlherms-ibm Please review as well

tgooding:
After discussion, we concluded that the umask settings defined in the systemctld service were the proper controls.  The umask is default is also correct - - we didn't want a user to mistakingly set file/directory ACLs with wide permissions on an ephemeral, single-user file system and then those permissions are blindly copied into cluster-wide storage.  System admins can override this behavior if they chose.  Closing PR.

PR #986, opened 2020-11-04
Do csm init of structures, use local string for config get vg name

meahoibm:
Unit tested that db entry has bb for volume group on IST systems f5n01 and f5n05.


tgooding:
I think you need a similar change on line 382.

bb/src/nodecontroller_csm.cc
```diff
@@ -178,7 +178,7 @@ NodeController_CSM::NodeController_CSM():
     vg.ssd_info[0]    = &ssdinfo;
 
     LOG(bb,info) << "Trying to create volume group:  ssd=" << nvmssd << "  vgfree=" << vgfree 
-                 << "  vgtotal=" << vgtotal << " volumeGroup="<< volumeGroup;
+                 << "  vgtotal=" << vgtotal << " volumeGroup="<< volumeGroup << "hostname=" << myhostname;
```
tgooding, line 181: You probably want a space before hostname.  e.g.,:  " hostname"

tgooding (approved):
Looks good - thanks

PR #989, opened 2020-11-13
Handle mtime conversion conditioned on getstats success

meahoibm:
Tested and works on IST system with ln01 launch node and compute node f5n05 using a bsub with stage in request.


tgooding (approved):
Looks good - thanks!

Issue #991, opened 2020-12-11
elastic search returns total_hits as dict object not integer


thanh-lam:
**Describe the bug**
A regression test on the CSM bigdata python command `"findJobTimeRange.py"` produced following message even though the allocation ID (1) in question did exist. That meant "total_hits" has the value 1.

```
[root@c650f99p06 python]# /opt/ibm/csm/bigdata/python/findJobTimeRange.py -a 1
# Found {'value': 1, 'relation': 'eq'} matches for specified the job.
# This implementation only supports queries where the hit count is equal to 1.
```

Adding a debug in the print statement in the script showed that total_hits has "value" as {'value': 1, 'relation': 'eq'} that also showed in message above.

```
[root@c685f4n07 python]# ./findJobTimeRange.py -j 114
# Found {'value': 1, 'relation': 'eq'} matches for specified the job.
# This implementation only supports queries where the hit count is equal to 1. Total hits=  {'value': 1, 'relation': 'eq'}
```

That led to the condition checking:
```
    if total_hits != 1:
        print("# This implementation only supports queries where the hit count is equal to 1.")
        return 3
```

Diagnosis: "total_hits" has "value" 1 but fell into the code path of `"total_hits != 1"`. So total_hits was not in the format of an integer 1. The output `{'value': 1, 'relation': 'eq'} hinted that it is a "dict" object. Further debugging confirmed that.

Python doesn't require data type declaration. Therefore, the change of total_hits from an integer to a "dict" object went without notice. But, the script failed the if condition due to the different format.

To adapt to this change of data type, change `total_hits` to `total_hits['value']`.

**To Reproduce**
1. Go to '/opt/ibm/csm/bigdata/python/'
2. Run ./findJobTimeRange.py -a 1      (make sure allocation 1 existed)
3. See message:
```
# Found {'value': 1, 'relation': 'eq'} matches for specified the job.
# This implementation only supports queries where the hit count is equal to 1.
```

**Expected behavior**
The command should display start and end time of the allocation. For example:
```
[root@c685f4n07 python]# ./findJobTimeRange.py -a 1
# Found {'value': 1, 'relation': 'eq'} matches for specified the job.allocation-id: 1
job-id: 526 - 0
user-name: root 
user-id: 0
begin-time: 2020-09-17.12:06:01:183 
end-time: 2020-09-17.12:06:58:382
```

**Screenshots**

**Environment (please complete the following information):**
 - Machine [CSM or IST CORAL cluster]
[...]

Issue #990, opened 2020-12-04
CSM BDS: python scripts TypeError with --state reverted


thanh-lam:
**Describe the bug**
For querying allocation data, CSM provides python scripts in /opt/ibm/csm/bigdata/python/. One example is "findUserJobs.py" that lists allocation info such as "state" and so on of a job. It produced following error when running with --state reverted. Other states (running, failed, complete) were listed with no error.

```
# ./findUserJobs.py -u tlam --state reverted
     State |   AID | P Job ID | S Job ID | Begin Time                 | End Time                  
Traceback (most recent call last):
  File "./findUserJobs.py", line 167, in <module>
    sys.exit(main(sys.argv))
  File "./findUserJobs.py", line 135, in main
    data.get("state")))
TypeError: unsupported format string passed to NoneType.__format__
```

**To Reproduce**
1. Login to CSM master or BDS node as root.
2. Change to /opt/ibm/csm/bigdata/python/ then run the command:
```
# ./findUserJobs.py -u <userid> --state reverted
```
3. See error.

**Expected behavior**
The command should not produce the error (which looked like an internal condition needed to be handled with the reverted state). Example of a good command output:
```
# ./findUserJobs.py -u root
     State |   AID | P Job ID | S Job ID | Begin Time                 | End Time                  
  complete |     1 |      526 | 0        | 2020-09-17 12:06:01.183794 | 2020-09-17 12:06:58.382907
  complete |     2 |      527 | 0        | 2020-09-17 12:06:02.483204 | 2020-09-17 12:06:56.800503
  complete |     3 |      528 | 0        | 2020-09-17 12:10:25.346929 | 2020-09-17 12:10:26.374556
  complete |     4 |      530 | 0        | 2020-09-17 12:12:52.830307 | 2020-09-17 12:12:53.838649
  complete |     5 |      531 | 0        | 2020-09-17 12:18:24.195737 | 2020-09-17 12:18:25.275275
  complete |     6 |      532 | 0        | 2020-09-17 12:20:02.431532 | 2020-09-17 12:20:27.420373
  complete |     7 |      533 | 0        | 2020-09-17 12:22:12.25542  | 2020-09-17 12:22:33.375314
  complete |     8 |      534 | 0        | 2020-09-17 12:27:51.261522 | 2020-09-17 12:28:11.101704
  complete |     9 |      535 | 0        | 2020-09-17 12:28:01.331114 | 2020-09-17 12:28:11.447308
[...]

PR #994, opened 2021-02-24
CSM BDS fixes related to open IST issues.

williammorrison2:
# Purpose
These fixes are related to issues #990 and #991 that Thanh and myself have been working on.
Examples and updates have been done in the issues tracker.

## TODOs
- [x] @williammorrison2 (check CSM FVT regression)
- [ ] @besawn  check the code.
- [ ] @thanh-lam  check the code.

### DIFF
--- a/csm_big_data/python/findJobsInRange.py
+++ b/csm_big_data/python/findJobsInRange.py
@@ -4,7 +4,7 @@
 #
 #    findJobsInRange.py
 #
-#    © Copyright IBM Corporation 2015-2020. All Rights Reserved
+#    © Copyright IBM Corporation 2015-2021. All Rights Reserved
 #
 #    This program is licensed under the terms of the Eclipse Public License
 #    v1.0 as published by the Eclipse Foundation and available at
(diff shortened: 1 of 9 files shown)


thanh-lam (approved):
For cast_helper.py, no change to
```
def convert_timestamp( timestamp ):
    ...
    return new_timestamp
```
because new_timestamp is a date in string format.

A note here about the change:
```
    if total_hits['value'] != 1:
        print("This implementation only supports queries where the hit count is equal to 1.")
        return 3
```

If may be more accurate with condition:
```
if total_hits['value'] > 1:
    ...
```

csm_big_data/python/findJobsInRange.py
```diff
@@ -120,9 +121,10 @@ def main(args):
         for hit in hits:
             data=cast.deep_get(hit, "_source", "data")
             if data:
+                condition = cast.deep_get(data, "history","end_time")
                 print(print_fmt.format(
                     data.get("allocation_id"), data.get("primary_job_id"), data.get("secondary_job_id"),
-                    data.get("begin_time"), cast.deep_get(data, "history","end_time"), 
+                    data.get("begin_time"), cast.deep_get(data, "history","end_time") if (condition!=None) else " ",
```
besawn, line 127: Suggestion for future improvement: try to deduce the number of calls to `cast.deep_get(data, "history","end_time")` from 2 to 1 if possible.

besawn (approved):

Issue #995, opened 2021-02-25
CSM BDS: findJobTimeRange.py is dependent on date+time format


thanh-lam:
**Describe the bug**
The python script `findJobTimeRange.py` displays `begin_time`, `history.end_time`, and related info for an allocation. As described in #992, the date+time format is required to be updated to "yyyy-MM-ddThh:mm:ss.SSSZ". The problem is that `findJobTimeRange.py` defined this date format based on the old date string "yyyy-MM-dd hh:mm:ss.SSSZ". This cause a failure:

```
[csm04][/u/tlam/git/CAST/csm_big_data/python]> ./findJobTimeRange.py -a 259
# Found 1 matches for specified the job.
Traceback (most recent call last):
  File "./findJobTimeRange.py", line 149, in <module>
    sys.exit(main(sys.argv))
  File "./findJobTimeRange.py", line 121, in main
    start_time=datetime.strptime(tr_data["begin_time"], '%Y-%m-%d %H:%M:%S.%f')
  File "/usr/lib64/python3.6/_strptime.py", line 565, in _strptime_datetime
    tt, fraction = _strptime(data_string, format)
  File "/usr/lib64/python3.6/_strptime.py", line 362, in _strptime
    (data_string, format))
ValueError: time data '2021-02-12T16:17:28.393028' does not match format '%Y-%m-%d %H:%M:%S.%f'
```

**To Reproduce**
1. Go to /opt/ibm/csm/bigdata/python
2. Run ./findJobTimeRange.py -a <ID>  #The allocation ID should exist
3. See error

**Expected behavior**
Example of a successful run:
```
[csm04][/u/tlam/git/CAST/csm_big_data/python]> ./findJobTimeRange.py -a 259
# Found 1 matches for specified the job.allocation-id: 259
job-id: 1146 - 0
user-name: tlam 
user-id: 11745
begin-time: 2021-02-12.16:17:28:393 
end-time: 2021-02-12.16:17:48:421
```

**Environment (please complete the following information):**
 - Machine [IST BDS cluster]
 - Version [CSM 1.2.8-3577]

**Additional context**
Following two lines in findJobTimeRange.py are where date format is defined.
```
        date_format= '%Y-%m-%d %H:%M:%S.%f'
...
        start_time=datetime.strptime(tr_data["begin_time"], '%Y-%m-%d %H:%M:%S.%f')
```

**Issue Source:**
Because the date format is "hard coded" in the script. It required to be changed every time the date+time format is changed.

[...]

Issue #992, opened 2021-01-12
CSM BDS python scripts don't work with --starttime or --endtime


thanh-lam:
**Describe the bug**
Regardless of the date entered with `--starttime` or `--endtime`, a script queries and returns all job or allocation records. The script `findUserJobs.py` is used here as example. All other scripts that support `--starttime` and `--endtime` options have similar issue.

**To Reproduce**
1. Go to /opt/ibm/csm/bigdata/python/
2. Run one of the scripts like `./findUserJobs.py --starttime 2021-01-12` to search for jobs started on that date.
3. See the command output with all jobs even at a different start time.
4. No error

**Expected behavior**
For these scripts to work correctly, they should only return the jobs within the time ranges specified with `--starttime` and|or `--endtime`.

**Screenshots**
```
[/opt/ibm/csm/bigdata/python]> export CAST_ELASTIC="...:9200"
[/opt/ibm/csm/bigdata/python]> ./findUserJobs.py -u tlam --starttime 2021-01-12
     State |   AID | P Job ID | S Job ID | Begin Time                 | End Time                  
    failed |    12 |      112 | 0        | 2020-09-23 15:33:19.868194 | 2020-09-23 15:34:16.817299
    failed |    13 |      113 | 0        | 2020-09-23 15:40:02.624777 | 2020-09-23 15:40:02.845004
    failed |    14 |      114 | 0        | 2020-09-23 15:44:55.221521 | 2020-09-23 15:44:55.431779
...
  complete |    96 |      195 | 0        | 2020-09-28 11:43:35.973273 | 2020-09-28 11:44:28.118919
  complete |    97 |      196 | 0        | 2020-09-28 12:42:20.720108 | 2020-09-28 12:42:21.852965
  complete |    98 |      197 | 0        | 2020-09-28 12:43:17.145406 | 2020-09-28 12:43:21.830506
  complete |    99 |      198 | 0        | 2020-09-28 12:44:37.372068 | 2020-09-28 12:44:38.463995
    failed |   100 |      299 | 0        | 2020-10-08 09:52:26.881603 | 2020-10-08 09:54:54.950164
```

I added some debugged prints in the script to show the time range that it created from the `--starttime` input.
```
[...]

PR #997, opened 2021-03-04
CSM BDS fixes related to date-time format issues #992 #995

thanh-lam:
# Purpose
Fixes for issues #992 #995.

# Approach
Updated logstash.conf file and modified python scripts in /opt/ibm/csm/csm_big_data/python/.

# Origin
#992
#995

# How to Test
For the logstash.conf file:
1. Copy CAST/csm_big_data/logstash/config/logstash.conf to /etc/logstash/conf.d/logstash.conf
2. Restart logstash: "systemctl restart logstash"

For the scripts, just try to run them in usual tests.

- [ ] Assign @williammorrison2 to review (So they can check regression)
- [ ] Assign 1 developer to @besawn

### DIFF
--- a/csm_big_data/logstash/config/logstash.conf
+++ b/csm_big_data/logstash/config/logstash.conf
@@ -94,6 +94,18 @@ else if "ras" in [tags] and "csm" in [tags] {
         add_field => { "type" => "log-ras"  } 
     }
 
+### Use logstash gsub filter plugin to process input data before sending to elastic index.
+### For the two fields: data.begin_time and data.history.end_time, the blank is replaced
+### with a letter T. Note: these fields have type "date" that requires the valid format.
+### Also note: No change to csm_transaction.log with this process.
+
+    mutate {
+        gsub => [
+           "[data][begin_time]", " ", "T",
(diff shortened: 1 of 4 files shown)


thanh-lam:
@besawn , I checked in an "urgent" fix late yesterday for changing TIME_SEARCH_FORMAT in `cast_helper.py`
```
#TIME_SEARCH_FORMAT = "epoch_millis"
TIME_SEARCH_FORMAT = "strict_date_optional_time_nanos"
```

This fixes the problem of Elasticsearch query by time range. Part of this time range query not working was due to the search time format in epoch_millis but the date+time format of "begin_time" and "end_time" are in nano seconds.

@williammorrison2 and I tested this fix yesterday and it worked very well. So we didn't want to miss this PR. Changing TIME_SEARCH_FORMAT also led to eliminating the call to `convert_time()`. Another effect of this fix is that: User has to input "complete" time format for `--starttime` and `--endtime`. For example:

"2021-03-08T16:00:00" instead of "2021-03-08T16".

I'll add comments into #992 to document this fix.

williammorrison2 (approved):
These files look good.

Issue #1003, opened 2021-04-30
Duplicate transfer request within the same job prevents cleanup of the job's logical volume


dlherms-ibm:
When submitting the same transfer definition twice within a job and also specifying the same handle and contribid, one of the transfers is successful and the other fails.  That is expected.  However, after both operations have run to completion, file locks remain on the compute node.  These file locks prevent the unmount of the path to the job's logical volume and the deletion of that logical volume.  The work around is to end bbProxy for the compute node, manually unmount the path to the logical volume, and then restart bbProxy.  Restarting bbProxy will automatically delete that orphaned logical volume.

While this is an application error that performs the same transfer twice using the same handle and contribid, the system should not be left in a state such that the job's logical volume cannot be properly cleaned up.

dlherms-ibm:
So, the problem is that when the second start transfer runs, it opens the source file, builds a file handle, and inserts it into the file handle registry overlaying the first file handle which then leaks the fd.  Then the failing second transfer removes the file handle it just inserted into the registry.  When the first transfer actually completes, it can't find any file handle to close upon completion of the transfer for the file.

I put in additional info logging for fh and fh registry logic in the following bbproxy log that shows what is happening.  Entry at timestamp 2021-04-29 13:10:25.802870 shows the overlay and the timestamp at 2021-04-29 13:10:28.938363 shows where we can't find the file handle upon completion of the actual transfer.

[Issue1003_Problem log.pdf](https://github.com/IBM/CAST/files/6406050/Issue1003_Problem.log.pdf)

Solution being pursued is to check to see if the same file handle entry exists in the registry prior to it being inserted.  If a duplicate, error out the operation at that point so that we do not leak the file handle and file descriptor.

PR #1004, opened 2021-04-30
Fix for issue #1003

dlherms-ibm:
This fix addresses issue #1003.  Solution is to prevent the add of a duplicate file handle in the registry.

Change to fh.cc has the change to the addFilehandle() method.  Previously, only a value of zero could be returned.  With this change, a -1 can now be returned indicating the duplicate file handle entry.  That -1 will be returned directly for bbProxy.  Due to restart logic, bbServer can actually run into this case when attempting to reuse a transfer definition that was already stored in the metadata.  In that case, we attempt to release the entry that already exists and then insert the new file handle.

There are four locations in the code that invoked addFilehandle(), three in LVUtils.cc and one in xfer.cc.  For LVUtils.cc, one location was adding the source-index to the file handle registry.  That was changed to receive the new return value.  The other two locations were for the target index and code just prior to the add was already checking for the duplicate.  No change for those two invocations.  For xfer.cc, it runs within bbServer and the change in fh.cc should suffice.

Another minor change was also made to fh.cc that now logs the correct fd value being closed.  Prior code always logged -1 as the fd being closed.

Prior code for bbProxy that ran when a start transfer failed was to close all file handles associated with the job, handle, and contribid.  Instead of unconditionally closing all, we now keep track of the file handles that this instance of start transfer has opened.  Upon failure, the code now only removes/closes those file handles that were opened for this instance of start transfer.

The following annotated bbProxy log now shows the original failing scenario as successful with the code changes.
[Issue1003_WithSolutionLog.pdf](https://github.com/IBM/CAST/files/6406438/Issue1003_WithSolutionLog.pdf)

### DIFF
--- a/bb/src/LVUtils.cc
+++ b/bb/src/LVUtils.cc
@@ -65,6 +65,8 @@ namespace bs  = boost::system;
 
 char MINIMUM_LOGICAL_VOLUME_SIZE[4] = "16M";
 
+typedef std::pair<uint32_t, filehandle*> FileHandleEntry;
+
 extern thread_local uid_t threadLocaluid;
 extern thread_local gid_t threadLocalgid;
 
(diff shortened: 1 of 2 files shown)


dlherms-ibm:
Pull request created, but additional testing is required...

tgooding (approved):

tgooding:
Looks good, merging.

PR #1006, opened 2021-06-17
For bbcmd OPEN_SERVER message to bbproxy, fix bbproxy connecting via SSL to named 

meahoibm:
Tested under bbrobot fvt and now works.
Improved ansible processing for updating burst buffer RPMs.
Track AUTHENTICATE message in logs.
LOG SSL_errror if it occurs on SSL_accept, SSL_connect, etc.

When processing OPEN_SERVER from bbcmd, bbproxy sleeps one second after SSL_connect to give SSL (re)negotiation to complete before sending AUTHENTICATE message to bbserver.  The SSL protocol and the application use the same socket for read/write.  The sleep(1) gives the open SSL protocol engine time to complete any hand shake between bbproxy and bbserver.  

Fix rare exception for a null pointer in connect.cc when a connection object is not returned because connect failed.

### DIFF
--- a/bb/src/connections.cc
+++ b/bb/src/connections.cc
@@ -1154,7 +1154,7 @@ int makeConnection2bbserver(const std::string& pName)
         return ENOTCONN;
     }
     addToConnectionMapsWithDoorbell(newconnection_sock);
-    sleep(1);//Wait a second for any outstanding SSL (re)negotiations to complete (avoid SSL read lockup)
+    sleep(1);//Wait a second for any outstanding SSL (re)negotiations to complete 
     std::string l_newconnection_name = newconnection_sock->getConnectName();
     int rc=xchgWithBBserver(l_newconnection_name);
     return rc;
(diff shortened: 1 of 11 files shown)


bb/src/connections.cc
```diff
@@ -1150,6 +1154,7 @@ int makeConnection2bbserver(const std::string& pName)
         return ENOTCONN;
     }
     addToConnectionMapsWithDoorbell(newconnection_sock);
+    sleep(1);//Wait a second for any outstanding SSL (re)negotiations to complete (avoid SSL read lockup)
```
tgooding, line 1157: Maybe add a comment in this issue on the need for the sleep(1).  From the comment, this is a workaround.  But a bit more explanation would be good and any alternatives.  For our current implementation using (largely) persistent connections, a 1 second connect delay isn't impactful.  Thanks.

tgooding (approved):
Looks good overall.  Thanks.

PR #1012, opened 2021-09-29
Bills read the docs bds cs21b

williammorrison2:


thanh-lam (commented):
A pending question about the Cast Search plugin for kibana that's supposed to be provided with ibm-csm-bds-kibana*rpm. But, this rpm is not available from the build. If this is true, then we'll need to add such a note in the Cast Search Section.

PR #1018, opened 2022-01-14
Add volumegroup and tcp transport options

tgooding:
Support bbactivate option to override LV volumegroup in the configuration file, and handle NVMf tcp transports


meahoibm:
Looks good.  Read surrounding code and looked at bb.cfg context.  Looked at nvme/tcp spec and looks a lot like nvme/rdma for parameters.

meahoibm (approved):

PR #1021, opened 2022-02-03
Update connections.cc

meahoibm:
Regression testing executed via robot.

### DIFF
--- a/bb/src/connections.cc
+++ b/bb/src/connections.cc
@@ -327,22 +327,22 @@ int setupWhoami(string whoami, string instance)
    \brief Connection authentication response handler
  */
 static boost::property_tree::ptree myV;
+static bool myV_set = false;
 boost::property_tree::ptree getVersionPropertyTree(){
+    if (!myV_set){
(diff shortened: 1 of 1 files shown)


bb/src/connections.cc
```diff
@@ -327,22 +327,22 @@ int setupWhoami(string whoami, string instance)
    \brief Connection authentication response handler
  */
 static boost::property_tree::ptree myV;
+static bool myV_set = false;
 boost::property_tree::ptree getVersionPropertyTree(){
+    if (!myV_set){
```
tgooding, line 332: Is this implicitly under the control of a global mutex?  Or is there a threading issue that might show up handling 50-100 simultaneous connections?  I'd suggest using a pthread_once() to control initialization (one thread enters, other threads block until 1st thread finishes).

meahoibm: Pushed up thread once change.  Tested with robot regression.

tgooding (approved):
lgtm

